PR #2, opened 2017-08-17
File

marcushines:
Update system set package to use streaming rpc
Update build files
Update gnoi.Path to compliant with gnmi.Path
Update simple test to use new path


types.proto
```diff
@@ -32,13 +32,25 @@ message HashType {
...
   bytes hash = 2;
 }
 
-// OpenConfig path representation of the target.
-// eg. /interfaces/interface[name=et-1/0/0]
+// Path encodes a data tree path as a series of repeated strings, with
+// each element of the path representing a data tree node name and the
+// associated attributes.
+// Reference: gNMI Specification Section 2.2.2.
 message Path {
```
robshakir: Is there a reason not to have a dependency between gNOI and gNMI here? Should we move to a common Path definition?

marcushines: yes we should but i didn't want to block on that as it is a simple change once we figure out what that means

PR #7, opened 2018-01-31
Fix references in gNOI protos; cleanup.

robshakir:
This CL fixes references to the openconfig/reference repo in the gNOI protobufs; regenerates the set of protos and performs some misc cleanup (add licensing to the `test` Go files; ensure that there is a package file for the tests; add a `compile_protos.sh` script).


test/simple_test.go
```diff
@@ -1,11 +1,25 @@
...
 package simple
 
 import (
 	"testing"
 
-	"github.com/golang/protobuf/proto/proto"
-	"github.com/openconfig/gnoi"
-	gbgp "github.com/openconfig/gnoi/bgp"
+	"github.com/golang/protobuf/proto"
+	gpb "github.com/openconfig/gnoi"
```
aashaikh: curious -- why do you need to have a label fo this instead of just using `gnoi` ?

robshakir: This is a style-guide rule - we generally always name things `pb` at the end to make it clear that they were generated by a protobuf. In packages that require only one proto, you might see it aliased to just `pb`.

robshakir:
Thanks for the review!

PR #8, opened 2018-04-06
SetPackage support for remote downloads

ejbrever:
@aashaikh @robshakir @samribeiro Can you please review?

### DIFF
--- a/system/system.proto
+++ b/system/system.proto
@@ -285,11 +285,11 @@ message RemoteDownload {
   string path = 1;
 
   enum DownloadType {
-    DL_UNKNOWN = 0;
-    DL_SFTP = 1;
-    DL_HTTP = 2;
-    DL_HTTPS = 3;
-    DL_SCP = 4;
+    UNKNOWN = 0;
+    SFTP = 1;
+    HTTP = 2;
+    HTTPS = 3;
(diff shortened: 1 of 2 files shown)


robshakir:
Looks good to me. Let’s ping @samribeiro on this. @aashaikh is OOO this week.

system/system.proto
```diff
@@ -275,11 +275,37 @@ message TracerouteResponse {
...
+// package from a remote location.
+message RemoteDownload {
+  // The path information containing where to download the package from.
+  // For HTTP(S), this will be the URL (i.e. foo.com/file.tbz2).
+  // For SFTP and SCP, this will be the address:/path/to/file
+  // (i.e. host.foo.com:/bar/baz).
+  string path = 1;
+
+  enum DownloadType {
+    DL_UNKNOWN = 0;
```
samribeiro: I see that RemoteDownload is only used as part of a Request and never a Response, so in what situations would you use DL_UNKNOWN?

samribeiro: I see that RemoteDownload is only used as part of Request and never a Response. In what situation would you use a DL_UNKNOWN value? Perhaps better to just use SFTP as default?

PR #9, opened 2018-05-22
First cut of Diag gNOI micro service

aghaffarkhah:
This PR defines a new gNOI micro service called Diag to perform diagnostic operations on a
switching box. At the moment the service provides APIs for BERT operation only. BURNIN will be added later. 

The APIs defined here are meant o deprecate the BERT-related APIs defined in gnoi/layer2/layer2.proto. The new APIs are more extensive than the old ones and match use cases in a real prod environment.

### DIFF
--- a/diag/diag.proto
+++ b/diag/diag.proto
@@ -23,19 +23,23 @@ package gnoi.diag;
 // The Diag service exports to main set of RPCs:
 // 1- BERT related RPCs: Used to perform Bit Error Rate Test (BERT)
 //    operations on a set of ports.
-// 2- BURNING related RPCs: Used to perform a vendor-provided Burnin test on the
+// 2- BURNIN related RPCs: Used to perform a vendor-provided Burnin test on the
 //    switch to ensure the box is ready to start serving traffic. Burnin tests
-//    are typically run in the datacenter, as part of turnup or repair workflow.
+//    are typically run in the field, as part of turnup or repair workflow.
 // Note: The RPCs defined here are are stateless operations and them
 // failing/passing should not leave any permanent artifact on the switch (unless
 // there is something wrong HW-wise).
+// Note: By "port" we refer to a channelized frontpanel or backplane port on a
+// chassis. In OpenConfig YANG models, there is a one-to-one relationship
+// between a port as used here and an "interface". Therefore, the "gnoi.Path"
+// fields below for ports correspond to "/interfaces/interface" in YANG models.
 service Diag {
   // Starts BERT operation on a set of ports. Each BERT operation is uniquely
   // identified by an ID, which is given by the caller. The caller can then
   // use this ID (as well as the list of the ports) to stop the BERT operation
   // and/or get the BERT results. This RPC is expected to return an error status
   // in the following situations:
-  // - When BERT operation is not supported on any of the ports specified by
+  // - When BERT operation is supported on none of the ports specified by
   //   the request.
   // - When BERT is already in progress on any port specified by the request.
   // - In case of any low-level HW/SW internal errors.
@@ -54,10 +58,10 @@ service Diag {
   // - When the BERT operation ID does not match the in progress or completed
   //   BERT operation on any of the ports specified by the request.
   // The RPC returns an OK status of none of these situations is encountered.
-  // Note that a BERT operation is considered completed if switch has a
-  // record of it. Also note that it is OK to receive a stop request for a port
-  // which has completed BERT, as long as the recorded BERT operation ID matches
-  // the one specified by the request.
+  // Note that a BERT operation is considered completed if the switch has a
+  // record/history of it. Also note that it is OK to receive a stop request for
+  // a port which has completed BERT, as long as the recorded BERT operation ID
+  // matches the one specified by the request.
   rpc StopBERT(StopBERTRequest) returns(StopBERTResponse) {}
 
   // Gets BERT results during the BERT operation or after it completes. The
@@ -113,10 +117,10 @@ enum BertStatus {
   BERT_STATUS_TEST_DURATION_TOO_LONG = 8;
   // The given BERT operation ID is not known. Returned for `StopBert` and
   // `GetBertResult` RPCs.
-  BERT_STATUS_BERT_OPERATION_ID_NOT_FOUND = 9;
+  BERT_STATUS_OPERATION_ID_NOT_FOUND = 9;
   // The given BERT operation ID is already in use. Returned when `StartBert`
   // RPC uses an ID which is already memorized for a BERT operation.
-  BERT_STATUS_BERT_OPERATION_ID_ALREADY_IN_USE = 10;
+  BERT_STATUS_OPERATION_ID_IN_USE = 10;
   // Failure to get the peer lock.
   BERT_STATUS_PEER_LOCK_FAILURE = 11;
   // Lost the peer lock after locking once.
@@ -133,8 +137,8 @@ message StartBERTRequest {
     gnoi.Path interface = 1;  // required
     // The selected PRBS generating polynomial for BERT.
     PrbsPolynomial prbs_polynomial = 2;  // required
-    // BERT duration in mins. Must be a positive number.
-    uint32 test_duration_in_mins = 3;  // required
+    // BERT duration in seconds. Must be a positive number.
+    uint32 test_duration_in_secs = 3;  // required
   }
   // Unique BERT operation ID specified by the client. Multiple BERTs run on
   // different ports can have the same BERT operation ID. This ID will be used
@@ -209,7 +213,7 @@ message GetBERTResultResponse {
     // Path to the interface corresponding to the port.
     gnoi.Path interface = 1;
     // BERT result get status for this port. Only if the status is
-    // BERT_STATUS_OK the rest of the fields below are meaningful.
+    // BERT_STATUS_OK are the rest of the fields meaningful.
     BertStatus status = 2;
     // The ID of the BERT operation running on this port. Since the caller
     // can query the BERT results for all the ports, ID can potentially be
--- a/diag/diag.proto
+++ b/diag/diag.proto
@@ -25,8 +25,10 @@ package gnoi.diag;
 //    operations on a set of ports.
 // 2- BURNING related RPCs: Used to perform a vendor-provided Burnin test on the
 //    switch to ensure the box is ready to start serving traffic. Burnin tests
-//    are typically run by HwOps in the datacenter, as part of turnup or repair
-//    workflow.
+//    are typically run in the datacenter, as part of turnup or repair workflow.
+// Note: The RPCs defined here are are stateless operations and them
+// failing/passing should not leave any permanent artifact on the switch (unless
+// there is something wrong HW-wise).
 service Diag {
   // Starts BERT operation on a set of ports. Each BERT operation is uniquely
   // identified by an ID, which is given by the caller. The caller can then
--- a/diag/diag.proto
+++ b/diag/diag.proto
@@ -22,8 +22,8 @@ package gnoi.diag;
 
 // The Diag service exports to main set of RPCs:
 // 1- BERT related RPCs: Used to perform Bit Error Rate Test (BERT)
-//    operation on a set of ports.
-// 2- BURNING related RPCs: Used to perform a vendor-provide Burnin test on the
+//    operations on a set of ports.
+// 2- BURNING related RPCs: Used to perform a vendor-provided Burnin test on the
 //    switch to ensure the box is ready to start serving traffic. Burnin tests
 //    are typically run by HwOps in the datacenter, as part of turnup or repair
 //    workflow.
@@ -32,7 +32,7 @@ service Diag {
   // identified by an ID, which is given by the caller. The caller can then
   // use this ID (as well as the list of the ports) to stop the BERT operation
   // and/or get the BERT results. This RPC is expected to return an error status
-  // in the the following situations:
+  // in the following situations:
   // - When BERT operation is not supported on any of the ports specified by
   //   the request.
   // - When BERT is already in progress on any port specified by the request.
@@ -130,7 +130,7 @@ message StartBERTRequest {
     // Path to the interface corresponding to the port.
     gnoi.Path interface = 1;  // required
     // The selected PRBS generating polynomial for BERT.
-    PrbsPolynomial prbs_polynomial = 2;
+    PrbsPolynomial prbs_polynomial = 2;  // required
     // BERT duration in mins. Must be a positive number.
     uint32 test_duration_in_mins = 3;  // required
   }
--- a/diag/diag.proto
+++ b/diag/diag.proto
@@ -15,7 +15,7 @@
 //
 
 // This file defines the gNOI APIs used to perform diagnostic operations on a
-// switching box.
+// network device.
 syntax = "proto3";
 
 package gnoi.diag;
@@ -24,11 +24,12 @@ package gnoi.diag;
 // 1- BERT related RPCs: Used to perform Bit Error Rate Test (BERT)
 //    operations on a set of ports.
 // 2- BURNIN related RPCs: Used to perform a vendor-provided Burnin test on the
-//    switch to ensure the box is ready to start serving traffic. Burnin tests
-//    are typically run in the field, as part of turnup or repair workflow.
+//    network device to ensure the device is ready to start serving traffic.
+//    Burnin tests are typically run in the field, as part of turnup or repair
+//    workflow.
 // Note: The RPCs defined here are are stateless operations and them
-// failing/passing should not leave any permanent artifact on the switch (unless
-// there is something wrong HW-wise).
+// failing/passing should not leave any permanent artifact on the network device
+//  (unless there is something wrong HW-wise).
 // Note: By "port" we refer to a channelized frontpanel or backplane port on a
 // chassis. In OpenConfig YANG models, there is a one-to-one relationship
 // between a port as used here and an "interface". Therefore, the "gnoi.Path"
@@ -58,7 +59,7 @@ service Diag {
   // - When the BERT operation ID does not match the in progress or completed
   //   BERT operation on any of the ports specified by the request.
   // The RPC returns an OK status of none of these situations is encountered.
-  // Note that a BERT operation is considered completed if the switch has a
+  // Note that a BERT operation is considered completed if the device has a
   // record/history of it. Also note that it is OK to receive a stop request for
   // a port which has completed BERT, as long as the recorded BERT operation ID
   // matches the one specified by the request.
@@ -66,7 +67,7 @@ service Diag {
 
   // Gets BERT results during the BERT operation or after it completes. The
   // caller uses the BERT operation ID it previously used when starting the
-  // operation to query it. The switch is expected to keep the results for
+  // operation to query it. The device is expected to keep the results for
   // last N BERT operations for some amount of time, as specified by the
   // product requirement. This RPC is expected to return error status in the
   // following situations:
@@ -75,7 +76,7 @@ service Diag {
   // - When the BERT operation ID does not match the in progress or completed
   //   BERT operation on any of the ports specified by the request.
   // The RPC returns an OK status of none of these situations is encountered.
-  // Note that a BERT operation is considered completed if switch has a
+  // Note that a BERT operation is considered completed if device has a
   // record of it.
   rpc GetBERTResult(GetBERTResultRequest) returns(GetBERTResultResponse) {}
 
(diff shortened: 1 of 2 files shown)


diag/diag.proto
```diff
@@ -0,0 +1,236 @@
...
+syntax = "proto3";
+
+package gnoi.diag;
+
+// The Diag service exports to main set of RPCs:
+// 1- BERT related RPCs: Used to perform Bit Error Rate Test (BERT)
+//    operations on a set of ports.
+// 2- BURNING related RPCs: Used to perform a vendor-provided Burnin test on the
+//    switch to ensure the box is ready to start serving traffic. Burnin tests
+//    are typically run by HwOps in the datacenter, as part of turnup or repair
```
ejbrever: I think we should remove this sentence about HwOps as this is Google specific. Different companies will do things differently.

diag/diag.proto
```diff
@@ -0,0 +1,236 @@
...
+//    workflow.
+service Diag {
```
ejbrever: We should probably note that these are only for stateless diagnostics activities (in other words, these activities will all end on their own). None of these should hold any state on the device. Stateful configuration should use the OpenConfig model with GNMI. I only mention this as other vendors support a stateful internal BERT and I am hoping we can expand the current OpenConfig models to fully support this. There is a test-signal boolean leaf defined at the moment though.

diag/diag.proto
```diff
@@ -0,0 +1,236 @@
...
+  // identified by an ID, which is given by the caller. The caller can then
+  // use this ID (as well as the list of the ports) to stop the BERT operation
+  // and/or get the BERT results. This RPC is expected to return an error status
+  // in the following situations:
+  // - When BERT operation is not supported on any of the ports specified by
+  //   the request.
+  // - When BERT is already in progress on any port specified by the request.
+  // - In case of any low-level HW/SW internal errors.
+  // The RPC returns an OK status of none of these situations is encountered.
+  rpc StartBERT(StartBERTRequest) returns(StartBERTResponse) {}
```
ejbrever: Curious on your thoughts, but in another PR, I am thinking we should make sure there is an OpenConfig state leaf to describe that an interface is being used for testing. Since this is a service affecting activity, we would want a way for streaming telemetry to indicate that even though the interface may be UP/UP, it is currently under test.

diag/diag.proto
aghaffarkhah: Done. Thanks for noticing.

diag/diag.proto
aghaffarkhah: oh definitely agree with that. This was actually in our todo. Since BERT operation are long lived operations, the oper state of an interface MUST be able to capture this. 

Internally we actually do have a specific state for a port when it is in so called "diag mode". We can definitely extend gnmi to capture that as well.

diag/diag.proto
aghaffarkhah: Sure these are stateless operations and them failing/passing should not have a permanent artifact on the switch (unless there is something wrong HW-wise). Added a point on that to the comments. 

However I am also interested to learn more about the statefull BERT you mentioned here. What is it and how/when it is used? Any pointers?

aghaffarkhah:
Addressed the comment. Thanks for the review.

diag/diag.proto
ejbrever: Here is a pointer to the leaf:
https://github.com/openconfig/public/blob/e8981a183157a5f730f3d054da63f9314f43efb4/release/models/optical-transport/openconfig-terminal-device.yang#L754

Essentially it is very similar, but the devices support permanently turning on a test signal which sends PRBS from the port. This is useful during capacity adds and sometimes troubleshooting to see if a link or portion of a link is error free. The optical vendors I have seen just do this as stateful config instead of running for specified time.

diag/diag.proto
```diff
@@ -0,0 +1,238 @@
...
+// This file defines the gNOI APIs used to perform diagnostic operations on a
+// switching box.
+syntax = "proto3";
+
+package gnoi.diag;
+
+// The Diag service exports to main set of RPCs:
+// 1- BERT related RPCs: Used to perform Bit Error Rate Test (BERT)
+//    operations on a set of ports.
+// 2- BURNING related RPCs: Used to perform a vendor-provided Burnin test on the
```
robshakir: s/BURNING/BURNIN/?

diag/diag.proto
```diff
@@ -0,0 +1,238 @@
...
+//    switch to ensure the box is ready to start serving traffic. Burnin tests
+//    are typically run in the datacenter, as part of turnup or repair workflow.
```
robshakir: Maybe say `typically run in the field`, since not all gNOI deployed devices will live in datacentres :-)

diag/diag.proto
```diff
@@ -0,0 +1,238 @@
...
+// Note: The RPCs defined here are are stateless operations and them
+// failing/passing should not leave any permanent artifact on the switch (unless
+// there is something wrong HW-wise).
+service Diag {
+  // Starts BERT operation on a set of ports. Each BERT operation is uniquely
+  // identified by an ID, which is given by the caller. The caller can then
+  // use this ID (as well as the list of the ports) to stop the BERT operation
+  // and/or get the BERT results. This RPC is expected to return an error status
+  // in the following situations:
+  // - When BERT operation is not supported on any of the ports specified by
```
robshakir: Does this mean "not supported on one or more of the ports", or "not supported on all the ports"? I assume the former, but it might be worth clarifying.

diag/diag.proto
robshakir: We actually have a `TESTING` state in `/interfaces/interface/state/oper-status` already. Would this be OK for this use case?

diag/diag.proto
```diff
@@ -0,0 +1,238 @@
...
+  // - When there is at least one BERT operation in progress on a port which
+  //   cannot be stopped in the middle of the operation (either due to lack of
+  //   support or internal problems).
+  // - When no BERT operation, which matches the given BERT operation ID, is in
+  //   progress or completed on any of the ports specified by the request.
+  // - When the BERT operation ID does not match the in progress or completed
+  //   BERT operation on any of the ports specified by the request.
+  // The RPC returns an OK status of none of these situations is encountered.
+  // Note that a BERT operation is considered completed if switch has a
+  // record of it. Also note that it is OK to receive a stop request for a port
```
robshakir: I'm not clear what "if switch has a record of it" means. Does this mean to say that only BERT operations that the switch has the history of are considered completed. For example, a BERT operation that ran before the last reboot might be unknown?

diag/diag.proto
```diff
@@ -0,0 +1,238 @@
...
+    // Path to the interface corresponding to the port.
+    gnoi.Path interface = 1;  // required
+    // The selected PRBS generating polynomial for BERT.
+    PrbsPolynomial prbs_polynomial = 2;  // required
+    // BERT duration in mins. Must be a positive number.
+    uint32 test_duration_in_mins = 3;  // required
+  }
+  // Unique BERT operation ID specified by the client. Multiple BERTs run on
+  // different ports can have the same BERT operation ID. This ID will be used
+  // later to stop the operation and/or get its results.
```
robshakir: I think we assume that the BERT operation ID is not unique to a client right? Maybe we should make this clear above. Your `BERT_STATUS_BERT_OPERATION_ID_ALREADY_IN_USE` seems to indicate this.

diag/diag.proto
```diff
@@ -0,0 +1,238 @@
...
+  // The specified test duration is too small.
+  BERT_STATUS_TEST_DURATION_TOO_SHORT = 7;
+  // The specified test duration is larger than maximum allowed.
+  BERT_STATUS_TEST_DURATION_TOO_LONG = 8;
+  // The given BERT operation ID is not known. Returned for `StopBert` and
+  // `GetBertResult` RPCs.
+  BERT_STATUS_BERT_OPERATION_ID_NOT_FOUND = 9;
+  // The given BERT operation ID is already in use. Returned when `StartBert`
+  // RPC uses an ID which is already memorized for a BERT operation.
+  BERT_STATUS_BERT_OPERATION_ID_ALREADY_IN_USE = 10;
```
robshakir: (nit) This seems quite long. Can we call this `BERT_STATUS_OPERATION_ID_IN_USE`?

diag/diag.proto
```diff
@@ -0,0 +1,238 @@
...
+  // There is no BERT running on the specified port. Returned when `StopBert`
+  // or `GetBertResult` RPC was called for an idle port.
+  BERT_STATUS_PORT_NOT_RUNNING_BERT = 6;
+  // The specified test duration is too small.
+  BERT_STATUS_TEST_DURATION_TOO_SHORT = 7;
+  // The specified test duration is larger than maximum allowed.
+  BERT_STATUS_TEST_DURATION_TOO_LONG = 8;
+  // The given BERT operation ID is not known. Returned for `StopBert` and
+  // `GetBertResult` RPCs.
+  BERT_STATUS_BERT_OPERATION_ID_NOT_FOUND = 9;
```
robshakir: `s/\_BERT\_//` perhaps? `BERT_STATUS_OPERATION_ID_NOT_FOUND` seems reasonable.

diag/diag.proto
```diff
@@ -0,0 +1,238 @@
...
+  // Failure to get the peer lock.
+  BERT_STATUS_PEER_LOCK_FAILURE = 11;
+  // Lost the peer lock after locking once.
+  BERT_STATUS_PEER_LOCK_LOST = 12;
+  // Misc internal errors that cannot be categorized by any of the previous
+  // error codes.
+  BERT_STATUS_INTERNAL_ERROR = 13;
+}
+
+message StartBERTRequest {
```
robshakir: Is there a reason to have this 1:N cardinality for the RPC to port relationship? It seems like we might simplify some things by having it be 1:1, at the cost of having multiple RPCs. The actual BERT on the port seems like it'd be disjoint either way.

diag/diag.proto
```diff
@@ -0,0 +1,238 @@
...
+  // Per port BERT start requests.
+  message PerPortRequest {
+    // Path to the interface corresponding to the port.
+    gnoi.Path interface = 1;  // required
+    // The selected PRBS generating polynomial for BERT.
+    PrbsPolynomial prbs_polynomial = 2;  // required
+    // BERT duration in mins. Must be a positive number.
+    uint32 test_duration_in_mins = 3;  // required
```
robshakir: I'd prefer we specify this in seconds -- is there a reason to only allow integer minutes?

diag/diag.proto
```diff
@@ -0,0 +1,238 @@
...
+  BERT_STATUS_INTERNAL_ERROR = 13;
+}
+
+message StartBERTRequest {
+  // Per port BERT start requests.
+  message PerPortRequest {
+    // Path to the interface corresponding to the port.
+    gnoi.Path interface = 1;  // required
+    // The selected PRBS generating polynomial for BERT.
+    PrbsPolynomial prbs_polynomial = 2;  // required
```
robshakir: Is it required to be able to run different polynomials per port? I'm not clear what the use case for this would be.

diag/diag.proto
```diff
@@ -0,0 +1,238 @@
...
+    // BERT start status for this port.
+    BertStatus status = 2;
+  }
+  // The same BERT operation ID given by the request.
+  string bert_operation_id = 1;
+  // Captures the results of starting BERT on a per-port basis.
+  repeated PerPortResponse per_port_responses = 2;
+}
+
+message StopBERTRequest {
```
robshakir: Here's where I think things get a bit more complicated by having multiple ports be part of the same BERT request. If I stop a BERT only on one port, but it has a single operation ID for that BERT, then this means that the BERT is cancelled for a subset of the ports, but running for the remainder. The current proto requires the caller to always check all statuses of the underlying ports to know the overall status. It seems like 1:1 would simplify this.

diag/diag.proto
```diff
@@ -0,0 +1,238 @@
...
+  bool result_from_all_ports = 3;
+}
+
+message GetBERTResultResponse {
+  // Per-port BERT results/status.
+  message PerPortResponse {
+    // Path to the interface corresponding to the port.
+    gnoi.Path interface = 1;
+    // BERT result get status for this port. Only if the status is
+    // BERT_STATUS_OK the rest of the fields below are meaningful.
```
robshakir: Suggest `Only if the status is BERT_STATUS_OK are the rest of the fields meaninful`

diag/diag.proto
```diff
@@ -0,0 +1,238 @@
...
+// 1- BERT related RPCs: Used to perform Bit Error Rate Test (BERT)
+//    operations on a set of ports.
+// 2- BURNING related RPCs: Used to perform a vendor-provided Burnin test on the
+//    switch to ensure the box is ready to start serving traffic. Burnin tests
+//    are typically run in the datacenter, as part of turnup or repair workflow.
+// Note: The RPCs defined here are are stateless operations and them
+// failing/passing should not leave any permanent artifact on the switch (unless
+// there is something wrong HW-wise).
+service Diag {
+  // Starts BERT operation on a set of ports. Each BERT operation is uniquely
```
robshakir: When we say `port` here, do we mean `interface` in OpenConfig parlance? Or do we mean a single physical port? I think it'd be good to clarify since it's not necessarily clear whether the `gnoi.Path` messages should correspond to `/interfaces/interface` entries or `/components/component` entries of type `PORT`.

diag/diag.proto
ejbrever: For optical devices, I think we also need this in the terminal-device/logical channels. We can expand this enum possibly:
https://github.com/openconfig/public/blob/e8981a183157a5f730f3d054da63f9314f43efb4/release/models/optical-transport/openconfig-terminal-device.yang#L767

aghaffarkhah: Thanks, for our use case TESTING state should be enough I think.

diag/diag.proto
aghaffarkhah: Done.

diag/diag.proto
aghaffarkhah: Done.

diag/diag.proto
aghaffarkhah: Thanks, clarified.

diag/diag.proto
aghaffarkhah: Yes you got it right. After a BERT operation is done, there is nothing in HW that keeps track of the completed BERTs. So it is up to the agent to keep track of them. The agent typically has a limited resource and keeps track of the last M BERTs. Also the storage used to keep track of BERTs might be volatile (wiped out after reboot). At any point of time if the switch knows about a previous BERT, it can consider this completed. Otherwise it is considered a new BERT. 

I added more details on this. PTAL.

diag/diag.proto
aghaffarkhah: No reason. The duration was given as mins before in CMAL case. But dont think it is a big deal to change the logic this much. Changed to seconds.

diag/diag.proto
aghaffarkhah: Chips support this. I see no harm in allowing this.

diag/diag.proto
aghaffarkhah: Done.

diag/diag.proto
aghaffarkhah: Ports refer to a channelized port which we previously identified using a (slot, port, channel). In YANG, I think an interface is a good equivalent. I clarified this in this text.

diag/diag.proto
aghaffarkhah: I strongly believe the capability of running BERT on multiple port in one RPC is required for tools. As also explained above. 

Regarding your example: 
The user can start BERT on multiple ports and put them under the same operation ID. That is true. He/she can also stop BERT on one or more ports which were part of the same operation ID later. The success is determines based on the overall BER values captured on all the ports which we get via GetBERTResult RPC. So if some of the per port BERTs were stopped earlier, things should still be clear. The BERT results will tell how long BERT were run on those ports and what was the BER value on those. To me this seems clear.

aghaffarkhah:
Addressed Rob's comments. PTAL.

diag/diag.proto
```diff
@@ -0,0 +1,242 @@
...
+    gnoi.Path interface = 1;  // required
+    // The selected PRBS generating polynomial for BERT.
+    PrbsPolynomial prbs_polynomial = 2;  // required
+    // BERT duration in seconds. Must be a positive number.
+    uint32 test_duration_in_secs = 3;  // required
+  }
+  // Unique BERT operation ID specified by the client. Multiple BERTs run on
+  // different ports can have the same BERT operation ID. This ID will be used
+  // later to stop the operation and/or get its results.
+  string bert_operation_id = 1;
```
aashaikh: might it be easier to manage these ids on the client if they are just fixed width numerics ?

diag/diag.proto
aashaikh: Couldn't we simplify the GetBERTResult by just having it return status for all the ports that were associated with the operation id (then don't need the two modes, all and per-port)

diag/diag.proto
```diff
@@ -0,0 +1,242 @@
...
+//    operations on a set of ports.
+// 2- BURNIN related RPCs: Used to perform a vendor-provided Burnin test on the
+//    switch to ensure the box is ready to start serving traffic. Burnin tests
+//    are typically run in the field, as part of turnup or repair workflow.
+// Note: The RPCs defined here are are stateless operations and them
+// failing/passing should not leave any permanent artifact on the switch (unless
+// there is something wrong HW-wise).
+// Note: By "port" we refer to a channelized frontpanel or backplane port on a
+// chassis. In OpenConfig YANG models, there is a one-to-one relationship
+// between a port as used here and an "interface". Therefore, the "gnoi.Path"
```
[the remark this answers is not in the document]
aghaffarkhah: Port is a well-known concept for BERT. I think we should mention it. Just referring to port as interface in the comment is going to be confusing too. 

Here we are just trying to make sure people will know what the ports refer to "when they think about YANG models".

robshakir:
This LGTM, thanks Ali. Please wait for Anees' comments before merging.

diag/diag.proto
```diff
@@ -0,0 +1,242 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+// This file defines the gNOI APIs used to perform diagnostic operations on a
+// switching box.
```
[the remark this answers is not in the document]
aghaffarkhah: I replace switching box and switch in general. 

What does the second comment mean though? Are you suggesting to provide a spec doc for BERT? Sure, but is that not good to have some explanation here too?

diag/diag.proto
aghaffarkhah: I agree with you. So far IDs have been used as strings in CMAL. Wanted to keep them consistent. But I like numerical IDs as well. 

For now I add a TODO and see if numbers are acceptable, if so I will change.

aghaffarkhah:
Hi Anees I tried to address your comments in the last commit. 

For some reason I could not comment on the last comment regarding GetBertResult. I am replying here. So as I understand it, the tools have a use case to get the results for all the ports that might be part of different operation IDs. 

As you said we might be able to simplify the logic a bit. I added a TODO there. Happy to simplify in case there is no use case.

diag/diag.proto
aghaffarkhah: So as I understand it, the tools have a use case to get the results for all the ports that might be part of different operation IDs.

As you said we might be able to simplify the logic a bit. I added a TODO there. Happy to simplify in case there is no use case.

diag/diag.proto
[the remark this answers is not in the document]
aashaikh: yes, but my view is that it's good to have some minimal explanation here and reference the spec for more detailed explanation (easier to consume in a doc IMO).  I think that change can be done in a subsequent PR.

diag/diag.proto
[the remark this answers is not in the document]
aashaikh: I just think it could be a little confusing since we specify which "ports" to test by specifying a path to an interface -- also, the port-to-interface relationship isn't 1:1 as described here for example (when you have channelized ports).

diag/diag.proto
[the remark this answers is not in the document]
aghaffarkhah: Having a detailed documentation SGTM. I will definitely take care of it. 

Note that this is just a beginning. I will still need to talk to many teams and do necessary changes.

diag/diag.proto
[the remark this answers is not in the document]
aghaffarkhah: I see what you are saying. Yeah I need to clarify this more. I will make sure to take care if this in the next PR.

PR #11, opened 2018-06-22
Support remote download of files when doing a File.Get

ejbrever:
@aashaikh @robshakir Can you please review this?

### DIFF
--- a/common.proto
+++ b/common.proto
@@ -1,0 +1,42 @@
+//
+// Copyright 2018 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+syntax = "proto3";
+
+package gnoi;
+
+import "third_party/openconfig/gnoi/types.proto";
+
+// RemoteDownload defines the details for a device to initiate a file transfer
+// from or to a remote location.
+message RemoteDownload {
+  // The path information containing where to download the data from or to.
+  // For HTTP(S), this will be the URL (i.e. foo.com/file.tbz2).
+  // For SFTP and SCP, this will be the address:/path/to/file
+  // (i.e. host.foo.com:/bar/baz).
+  string path = 1;
+
+  enum Protocol {
+    UNKNOWN = 0;
+    SFTP = 1;
+    HTTP = 2;
+    HTTPS = 3;
+    SCP = 4;
+  }
+  Protocol protocol = 2;
+
+  gnoi.Credentials credentials = 3;
+}
--- a/common.proto
+++ b/common.proto
@@ -18,7 +18,7 @@ syntax = "proto3";
 
 package gnoi;
 
-import "third_party/openconfig/gnoi/types.proto";
+import "github.com/openconfig/gnoi/types.proto";
 
 // RemoteDownload defines the details for a device to initiate a file transfer
 // from or to a remote location.
--- a/file/file.proto
+++ b/file/file.proto
@@ -18,11 +18,11 @@ syntax = "proto3";
 
 package gnoi.file;
 
-import "third_party/openconfig/gnoi/common.proto";
+import "github.com/openconfig/gnoi/common.proto";
 import "github.com/openconfig/gnoi/types.proto";
 
 service File {
-  // Get read and streams the contents of a file from the target.
+  // Get reads and streams the contents of a file from the target.
   // The file is streamed by sequential messages, each containing up to
   // 64KB of data. A final message is sent prior to closing the stream
   // that contains the hash of the data sent. An error is returned
--- a/file/file.proto
+++ b/file/file.proto
@@ -104,11 +104,11 @@ message GetResponse {
   }
 }
 
-// A GetRemoteRequest specifies the remote_file to transfer and the details on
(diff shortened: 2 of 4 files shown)


common.proto
```diff
@@ -0,0 +1,42 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+syntax = "proto3";
+
+package gnoi;
+
+import "third_party/openconfig/gnoi/types.proto";
```
aashaikh: import path cleanup

file/file.proto
```diff
@@ -28,6 +29,13 @@ service File {
   // if the file does not exist or there was an error reading the file.
```
aashaikh: "Get reads and streams ..."

system/system.proto
```diff
@@ -305,7 +285,7 @@ message Package {
   // For system packages this will take effect after reboot.
   bool activate = 5;
```
aashaikh: I still wonder if `activate` is the right name for this field -- to me a true value here could be interpreted as "make this package active now".  does something like `requires_reboot` make sense ?

common.proto
ejbrever: oops, thanks.

file/file.proto
```diff
@@ -18,6 +18,7 @@ syntax = "proto3";
 
 package gnoi.file;
 
+import "third_party/openconfig/gnoi/common.proto";
```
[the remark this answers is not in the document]
ejbrever: thanks :)

file/file.proto
ejbrever: done

system/system.proto
```diff
@@ -24,6 +24,7 @@ syntax = "proto3";
 
 package gnoi.system;
 
+import "third_party/openconfig/gnoi/common.proto";
```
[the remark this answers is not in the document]
ejbrever: Fixed

system/system.proto
ejbrever: I am open to a new name, just not sure what that might be :)

For more context, when you transfer a file, the file can be stored on the device, but not applied. So after transferring a file (without activation) a reboot of the device would not set the package to active. You have to specify somehow to set a specific package active.

file/file.proto
```diff
@@ -96,6 +104,21 @@ message GetResponse {
   }
 }
 
+// A GetRemoteRequest specifies the remote_file to transfer and the details on
+// where to transfer the data. The remote_file must be an absolute path to a
+// file.
+message GetRemoteRequest {
+  string remote_file = 1;
```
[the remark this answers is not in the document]
ejbrever: Yea, agree this is a bit confusing and the docs don't help :)

The intention was RemoteDownload specified how to get the file, and this was supposed to specify the path name in the device to save it to (i.e. /tmp/os-1.0).

Maybe local_file be a better name and I can update the descriptions as well. Please let me know what you think.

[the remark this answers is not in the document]
robshakir: Yep - something like `local_path` would probably make more sense here IMHO.

system/system.proto
aashaikh: I think I'm getting hung up on the "reboot" part of this -- maybe we can change the comment to say something like:
```
Indicates that the package should be made active after receipt on the device.
For system image packages, the new image is expected to be active after a reboot.
```

common.proto
```diff
@@ -0,0 +1,42 @@
...
+
+syntax = "proto3";
+
+package gnoi;
+
+import "github.com/openconfig/gnoi/types.proto";
+
+// RemoteDownload defines the details for a device to initiate a file transfer
+// from or to a remote location.
+message RemoteDownload {
```
aghaffarkhah: Sorry if I missed this, but why do you think this message is common to all the services? What is the reason behind creating this file? Sorry this is just me understanding...

file/file.proto
```diff
@@ -18,16 +18,24 @@ syntax = "proto3";
...
   // that contains the hash of the data sent. An error is returned
   // if the file does not exist or there was an error reading the file.
   rpc Get(GetRequest) returns (stream GetResponse) {}
 
+  // GetRemote transfers the contents of a file from the target to a specified
+  // remote location. The response contains the hash of the data transferred.
+  // An error is returned if the file does not exist, the file transfer fails,
+  // or if there was an error reading the file. This is a blocking call until
+  // the file transfer is complete.
+  rpc GetRemote(GetRemoteRequest) returns (GetRemoteResponse) {}
```
aghaffarkhah: Wouldnt "TransferToRemote" or something similar a better name? Sorry not trying to be obsessed with the name, but having "Get" in the name of the RPC is a bit confusing.

common.proto
ejbrever: No problem at all. It was originally in system.proto. In this change we wanted to add it to file.proto as well so it made sense to put it in a common place.

file/file.proto
ejbrever: I'm pretty open to a different name. Maybe TransferFromRemote would be more obvious here?

@aashaikh @robshakir Any thoughts on the name here as well? I know we had discussed this previously as well.

system/system.proto
ejbrever: Sure, done.

file/file.proto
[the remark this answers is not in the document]
ejbrever: Sure, done.

file/file.proto
robshakir: I'm fine with `TransferFromRemote` or even `FetchRemote` here, but I don't have strong feelings. 

The "get" in `GetRemote` is essentially "get this file from a remote place" - maybe fetch is better than "get" here?

ejbrever: I'm leaning towards TransferFromRemote, but please let me know if you feel otherwise.

ejbrever: After thinking again about it, I agree with @aghaffarkhah on TransferToRemote :)

I have updated using this name now. I think this is much more obvious than GetRemote.

ejbrever:
@robshakir @aghaffarkhah Can you please take another look?

Thanks,
Eric

PR #14, opened 2018-07-07
Some protobuf housekeeping

bocon13:
- Updating README to recommend `compile_protos.sh` to generate Go files
- Adding `diag` service to `compile_protos.sh`
- Adding explicit `go_package` and making proto imports relative

The last item is important because for non-Go languages, users will likely not have the fully qualified path including `github.com/openconfig/gnoi` as part of their local workspace.


bocon13:
@robshakir @aghaffarkhah  Does this break anything?

PR #15, opened 2018-10-01
Add a simplified security model

samribeiro:
In addition, allow to optionally install CA certificates using the gNOI Cert interface.


robshakir:
LGTM, thanks for the work Sam.

robshakir:
Ah, apologies. I sent you an invitiation to join the committers team.

PR #17, opened 2019-04-02
gNOI Factory Reset - service for factory resetting a Target

samribeiro:


ejbrever:
I think the Reboot RPC already supports this (although in a bit more rudimentary way):
https://github.com/openconfig/gnoi/blob/master/system/system.proto#L127

Would it make sense to enhance that RPC instead?

samribeiro:
@ejbrever I agree with coalescing, however I am reluctant about using the System service at all. In this use case for instance, it is inconvenient to ask that the Reboot RPC in System service be implemented, but no other RPC. That is at the moment our only requirement. It is hard to state requirements when a Service is too overarching, I believe we should aim for slimmer micro-services. Given this, how about extruding the Reboot RPCs from System and augmenting the Reset service?

samribeiro:
Thank you @ejbrever, I agree with you:
- I have updated the naming as you suggest in 2). 
- I have updated the System messaging as you suggest in 1).

ejbrever:
LGTM, thanks for making the changes!

factory_reset/reset.proto
```diff
@@ -0,0 +1,41 @@
...
+  // from factory.
+  bool factory_os = 1;
+  // Instructs the Target to zero fill persistent storage state data.
+  bool zero_fill = 2;
+}
+
+message InitiateResponse {
+  // Must be set when factory_os is set and the Target does not support it.
+  // Does not halt the Factory Reset process.
+  bool warning_unsupported_factory_os = 1;
```
aashaikh: Is it more canonical to return an error code (e.g., `INVALID_ARGUMENT`) and give this sort of information in the error message / details ?

system/system.proto
```diff
@@ -124,7 +124,7 @@ enum RebootMethod {
   HALT = 3;        // Halt, if possible.
   WARM = 4;        // Reload configuration but not underlying hardware.
   NSF = 5;         // Non-stop-forwarding reboot, if possible.
-  RESET = 6;       // Reboot to factory defaults.
+  reserved 6;      // Please use instead gNOI FactoryReset.Initiate().
```
aashaikh: maybe something like `RESET method is deprecated in favor of the gNOI FactoryReset.Initiate()`

system/system.proto
```diff
@@ -27,7 +27,7 @@ package gnoi.system;
 import "github.com/openconfig/gnoi/common/common.proto";
 import "github.com/openconfig/gnoi/types/types.proto";
 
-option (types.gnoi_version) = "0.1.0";
+option (types.gnoi_version) = "0.1.1";
```
aashaikh: I think this change should increment the minor version .

If there are implementations of gNOI System, we might even want to set to 1.0 per the OpenConfig convention.

factory_reset/reset.proto
```diff
@@ -0,0 +1,41 @@
...
+  // It is optional for the Target to support rolling back the OS to the same
+  // version shipped from factory. If not supported, the running OS should be
+  // used.
+  //
+  // It is optional for the Target to zero fill permanent storage where state
+  // data is stored.
+  rpc Initiate(InitiateRequest) returns (InitiateResponse);
+}
+
+message InitiateRequest {
```
aashaikh: are you interested in having a scheduled version where reset does not happen immediately?

system/system.proto
samribeiro: Thanks!

system/system.proto
samribeiro: Agree, lets then follow the convention.

factory_reset/reset.proto
samribeiro: The reasoning is that if a client wants to factory reset a Target, neither the fact that it does not return to factory os, or perform zero fill, would be reason to not proceed with the Factory Reset. Additionally a client can blanket Reset RPC a series of different Targets without having to store intent about zero fill or factory os capabilities. It will however have that information upon every Reset RPC. I am not incredibly strong about this, it just seems to me the least complex option given the use case.

factory_reset/reset.proto
```diff
@@ -0,0 +1,41 @@
...
+  // shipped from factory. State includes storage, configuration, logs,
+  // certificates and licenses.
+  //
+  // It is optional for the Target to support rolling back the OS to the same
+  // version shipped from factory. If not supported, the running OS should be
+  // used.
+  //
+  // It is optional for the Target to zero fill permanent storage where state
+  // data is stored.
+  rpc Initiate(InitiateRequest) returns (InitiateResponse);
```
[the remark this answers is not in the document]
samribeiro: Indeed, thanks!

factory_reset/reset.proto
```diff
@@ -0,0 +1,41 @@
...
+
+// The FactoryReset service exported by Targets.
+service FactoryReset {
+  // The Initiate RPC allows the Client to instruct the Target to immediately
+  // clean all existing state and boot the Target in the same condition as it is
+  // shipped from factory. State includes storage, configuration, logs,
+  // certificates and licenses.
+  //
+  // It is optional for the Target to support rolling back the OS to the same
+  // version shipped from factory. If not supported, the running OS should be
```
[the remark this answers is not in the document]
samribeiro: Please see the reasoning in the comment below.

factory_reset/reset.proto
samribeiro: It is a good idea, however I cannot come up with a use case for it. =)

factory_reset/reset.proto
aashaikh: @gcsl PTAL (and add refs to canonical error handling please)

gcsl: I suggest returning an explicit error if the device doesn't support the requested options and not proceed with a reset.  gRPC does allow returning a proto message as part of the error in the "details" field as Anees indicated, which could include the current message with booleans to indicate what wasn't accepted.

https://cloud.google.com/apis/design/errors#error_model

samribeiro: I have updated the service with documentation that states the Target should return gRPC error code UNIMPLEMENTED when either factory_os or zero_fill are requested but not implemented.

factory_reset/reset.proto
```diff
@@ -14,28 +14,22 @@ service FactoryReset {
...
-  // It is optional for the Target to zero fill permanent storage where state
+  // Optionally allows for the Target to zero-fill permanent storage where state
   // data is stored.
   rpc Start(StartRequest) returns (StartResponse);
 }
 
 message StartRequest {
   // Instructs the Target to rollback the OS to the same version as it shipped
-  // from factory.
+  // from factory. Target returns grpc status code UNIMPLEMENTED if set but not
```
gcsl: Did you consider INVALID_ARGUMENT so a caller could easily differentiate from the whole RPC not being implemented? The gRPC libraries can return UNIMPLEMENTED for certain conditions but will never return INVALID_ARGUMENT according to https://github.com/grpc/grpc/blob/master/doc/statuscodes.md

samribeiro: Thanks! Updated.

I wonder if it is not less ambiguous and more precise to define error codes in the service directly. In practice a Client will perform an RPC call with two of the not implemented bits set. It will receive a gRPC error with INVALID_ARGUMENT and some arbitrary string. It will need to then parse that string to figure out which one of the optional arguments is not supported and try again with the given option unselected. A Client will then be at the mercy of the implementation done by the Target. This problem would not occur If this is done with explicit error codes defined in the service, example:
https://github.com/openconfig/gnoi/blob/master/os/os.proto#L213

In general I am in favour of precise services that reduce guessing at the client implementation.

gcsl: I am in full agreement on having explicit error conditions.  My original concern was with the semantic of performing the reset even if one of the requested options was unsupported. I think it would be good to explicitly return an error at the gRPC layer and if there are multiple differentiable invalid arguments that might need to be handled by the client in different ways also include the response message as before.

samribeiro: so it turns out that if gRPC returns anything than !OK, gRPC does not guarantee the delivery of the RPC response message.
https://github.com/grpc/grpc/issues/16786#issuecomment-437878961

This is a generalization that will depend on the !OK type and the implementation, however it feels significant enougth to introduce ambiguity. And to avoid it, in order to use gRPC errors, only code and status should be used to provide extra information about the error. This brings ofcouse an assured loss of programatic detail about the nature of the error.

With this I am inclined more and more to not throwing a gRPC error and keeping the Error service specific. Made a new commit to reflect this. What are your thoughts?

samribeiro: I have updated the spec to return gRPC Status error code if any of the optional functionalities are not supported/implemented.

factory_reset/reset.proto
```diff
@@ -0,0 +1,56 @@
...
+  // Instructs the Target to rollback the OS to the same version as it shipped
+  // from factory.
+  bool factory_os = 1;
+  // Instructs the Target to zero fill persistent storage state data.
+  bool zero_fill = 2;
+}
+
+message ResetSuccess {}
+
+message ResetError {
```
aashaikh: May want to add a comment above this message to indicate this is the proto message that should be returned in the gRPC Status.details field.

samribeiro: thanks, done!

PR #18, opened 2019-04-02
Bootstrapping gNOI/gNMI using the gNOI cert service

samribeiro:


docs/simplified_bootstrapping.md
```diff
@@ -0,0 +1,111 @@
...
+such, in the default state, the Target should have and use a default
+certificate. This certificate should be used only in the default state.
+Resetting the Target to the default state always generates a new default
+certificate or reuses a default one.
+
+##### Certificate ID
+
+The first installed certificate and CA certificate pool is used for the gNOI and
+gNMI services. No other certificates should be installed during the default
+stage. If the certificate ID is specified, then that becomes the associated
```
Reshad-Rahman: This association is implicit. Why not do an explicit association by using the certificate-id leaf of openconfig-system-management.yang? I thought that was the intent (based on previous discussions).

docs/simplified_bootstrapping.md
```diff
@@ -0,0 +1,111 @@
...
+certificate ID for the gNMI/gNOI service. If none is specified then one is
+assigned by the Target.
+
+##### Target behaviour
+
+The Target must restart its gNOI/gNMI service in secure mode as soon as the
```
Reshad-Rahman: secure mode is applicable to the gNMI/gNOI service only if client authentication has been enabled via the ca_certificates bundle?

docs/simplified_bootstrapping.md
samribeiro: Reshad, that is a very good point. The guiding reason here is security. The gNOI/gNMI endpoint is at this stage accepting any connection, so it is at a securely fragile stage. The expectation here is to make sure that the only gNOI/gNMI action available is to install a certificate. This reduces the number of vectors an attacker has for meddling with the Target otherwise.

Please let me know what concerns you have with this approach and if you suggest an alternative.

docs/simplified_bootstrapping.md
samribeiro: Yes, however the wording here is not the best. There is no definition of "secure mode" in this document so its use is ambiguous. What it should say is:
"The Target must restart its gNOI/gNMI service and exclusively perform secure connections as soon as the...." Will send a Commit. Thanks for the highlight.

docs/simplified_bootstrapping.md
Reshad-Rahman: Sam, I now understand the motivation for doing this. The concern I had was because there's a well-defined explicit way, I prefer to avoid multiple ways of doing the same operation. But I get it now, and no I don't have an alternative.

docs/simplified_bootstrapping.md
Reshad-Rahman: Sounds good.

docs/simplified_bootstrapping.md
```diff
@@ -0,0 +1,112 @@
...
+Resetting the Target to the default state always generates a new default
+certificate or reuses a default one.
+
+##### Certificate ID
+
+The first installed certificate and CA certificate pool is used for the gNOI and
+gNMI services. No other certificates should be installed during the default
+stage. If the certificate ID is specified, then that becomes the associated
+certificate ID for the gNMI/gNOI service. If none is specified then one is
+assigned by the Target.
```
Reshad-Rahman: Sam, what is the motivation for not providing a certificate-id? If the target assigns one, how will the client know which one got assigned by the target? And if the client does not know the certificate-id, how can the client do Rotate or Revoke.

samribeiro: The motivation was to allow the platform to use its own default CertID string for this. The Client could then identify the certificate ID by performing a GetCertificatesRequest.
However, I am now realizing that this is adding complexity, it is in addition breaking the cert Install spec where a CertID is required. This is therefore unnecessary. Thank you for pointing it out. I am updating this PR to correct this.

PR #22, opened 2019-05-06
Suggest order in gNOI cert ca_certificates

samribeiro:
Set expectations in the order that CA certificates should have. This helps with improving performance in the Target when building chained relationships between certificates. This is merely suggested. Certificates can still be provided out of order.

### DIFF
--- a/cert/cert.proto
+++ b/cert/cert.proto
@@ -282,6 +282,8 @@ message LoadCertificateRequest {
   // Groups of chained certificates should be last, where within, the root
   // certificate is the last one. E.g.:
   // CertA, CertB, CertB-Root, CertC, CertC-Intermediate, CertC-Root
+  // Note that on a forward looking iteration, CA Certificates will be managed
+  // by a dedicated gNOI service.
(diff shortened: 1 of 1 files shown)


cert/cert.proto
```diff
@@ -278,6 +278,10 @@ message LoadCertificateRequest {
   // certificates should squash the existing bundle. This field provides a
   // simplified means to provision a CA bundle that can be used to validate
   // other peer's certificates.
+  // To improve performance in the Target, certificates can be ordered.
+  // Groups of chained certificates should be last, where within, the root
+  // certificate is the last one. E.g.:
+  // CertA, CertB, CertB-Root, CertC, CertC-Intermediate, CertC-Root
```
Reshad-Rahman: Sam, looks good. Please note there could also be CA-Cert for target certificate in the bundle.

samribeiro: Thanks Reshad.
Indeed, I have created another PR to cover that under the simplified security model document.

PR #30, opened 2019-11-21
PSD to Spectrum Power units change

ejbrever:
@aashaikh Can you please review? This is a continuation of the model change which moved away from the PSD units. Thanks!


ejbrever:
@googlebot I signed it!

wavelength_router/wavelength_router.proto
```diff
@@ -23,7 +23,7 @@ import "github.com/openconfig/gnoi/types/types.proto";
 option (types.gnoi_version) = "0.1.0";
```
aashaikh: this version will need to be updated -- do we know if there are implementations of this RPC already?

wavelength_router/wavelength_router.proto
```diff
@@ -34,17 +34,17 @@ service WavelengthRouter {
   // the initiating client. A client that becomes disconnected during an
   // operation can reconnect, requesting the same operation, and receive status
   // updates this way. There is no means by which an adjustment that has been
-  // triggered by a client can be cancelled unless the CancelAdjustPSD RPC is
-  // used.
-  rpc AdjustPSD(AdjustPSDRequest) returns (stream AdjustPSDResponse) {}
+  // triggered by a client can be cancelled unless the CancelAdjustSpectrum
+  // RPC is used.
+  rpc AdjustSpectrum(AdjustSpectrumRequest) returns (stream AdjustSpectrumResponse) {}
```
aashaikh: Do we need to consider add a new rpc rather than just replacing this one?  (is there an easier migration path?)

wavelength_router/wavelength_router.proto
ejbrever: Nope, fortunately, we caught this early on. With some luck, there will be implementations in 2020 :)

wavelength_router/wavelength_router.proto
ejbrever: Since there are not yet any implementations, I think this will be the cleanest way long term.

wavelength_router/wavelength_router.proto
ejbrever: Version is now updated. Can you please take another look?

wavelength_router/wavelength_router.proto
aashaikh: After some additional discussion, I think we should consider adding a new RPC rather than just wholesale replacing.  We can mark the existing rpc or message fields as deprecated, and remove it in a subsequent revision.

ejbrever: Sure, the RPCs are marked as deprecated. I can remove in a subsequent revision. Please let me know if this needs anything else.

wavelength_router/wavelength_router.proto
```diff
@@ -36,12 +36,38 @@ service WavelengthRouter {
...
+
+  // CancelAdjustSpectrum cancels an in-progress AdjustSpectrum request. If an
+  // adjustment is not being performed for the provided request, then the RPC
+  // should return successfully.
+  rpc CancelAdjustSpectrum(AdjustSpectrumRequest) returns (CancelAdjustSpectrumResponse) {}
 }
 
 message AdjustPSDRequest {
```
aashaikh: Do we need to mark these messages deprecated as well ?  What is the proto best practice?

wavelength_router/wavelength_router.proto
```diff
@@ -101,3 +127,60 @@ message AdjustPSDResponse {
 
 message CancelAdjustPSDResponse {}
 
+message AdjustSpectrumRequest {
+  // Absolute path to the /components/component list member in the OpenConfig
+  // model corresponding to the port to perform the adjustment for.
+  // (e.g., /components/component[name=amp-1-1])
```
aashaikh: wouldn't this be a port on the ROADM (not AMP?) in the case where they are combined, are the ports named for amplifier?

ejbrever: The important part is to indicate directionality to the system as these adjustments are made on a per direction basis. Some vendors use the amp port as a convenient way, but I agree, it logically makes more sense in the example for this to be a waverouter port (it also aligns well with the name of the proto :)

wavelength_router/wavelength_router.proto
ejbrever: I learned something new that a message can be marked deprecated :)

PR #32, opened 2020-01-15
compile_protos.sh: explicit bash shebang

alrs:
`/bin/sh` on my Debian Linux machine points to `/bin/dash` instead of `/bin/bash`, which breaks the `compile_protos.sh` script due to lack of support for `pipefail` in `dash`.

This sets the shebang to `bash`.


alrs:
A ping for @robshakir.

PR #44, opened 2020-11-25
Add intro document for authn/authz discussions.

robshakir:
```
 * (A) auth/authn-authz-intro.md
  - Add a document that describes the problem statement and proposed
    development approach for authn and authz for devices that
    are managed according to OpenConfig APIs (e.g., gNOI, gNMI, gRIBI)
    - based on discussion with the contributors.
```

@danielywong @johnwestbrook @aashaikh - PTAL.

### DIFF
--- a/auth/authn-authz-intro.md
+++ b/auth/authn-authz-intro.md
@@ -4,34 +4,81 @@
 **Published:** November 2020
 
 ## Problem Statement
-Historically, network devices have had a single store of configuration data, users with sufficient privilege levels are allowed to edit this data. Implementation-specific mechanisms have been used to restrict the subsets of configuration and telemetry that a particular user can interact with, such that it is possible to enforce particular roles interacting with a network device in specific manners. Often, the authorization policies that network operators have implemented have been coarse - for example, high privilege users being able to edit any configuration, or a single SNMP community having full access to all telemetry on the device. Typically, only lawful intercept credentials and configuration has been subject to a 'protected' configuration store, which was treated differently by the device.
+Historically, network devices have had a single store of configuration data,
+users with sufficient privilege levels are allowed to edit this data.
+Implementation-specific mechanisms have been used to restrict the subsets of
+configuration and telemetry that a particular user can interact with, such that
+it is possible to enforce particular roles interacting with a network device in
+specific manners. Often, the authorization policies that network operators have
+implemented have been coarse - for example, high privilege users being able to
+edit any configuration, or a single SNMP community having full access to all
+telemetry on the device. Typically, only lawful intercept credentials and
+configuration has been subject to a 'protected' configuration store, which was
+treated differently by the device.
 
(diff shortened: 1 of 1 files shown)


auth/authn-authz-intro.md
```diff
@@ -0,0 +1,37 @@
...
+**Published:** November 2020
+
+## Problem Statement
+Historically, network devices have had a single store of configuration data, users with sufficient privilege levels are allowed to edit this data. Implementation-specific mechanisms have been used to restrict the subsets of configuration and telemetry that a particular user can interact with, such that it is possible to enforce particular roles interacting with a network device in specific manners. Often, the authorization policies that network operators have implemented have been coarse - for example, high privilege users being able to edit any configuration, or a single SNMP community having full access to all telemetry on the device. Typically, only lawful intercept credentials and configuration has been subject to a 'protected' configuration store, which was treated differently by the device.
+
+The OpenConfig project initially provided means to inter-work with existing means of authorization and authentication on network devices - providing means to supply a username and password via gRPC metadata for gNMI and gNOI - along with mutual TLS authentication. However, this approach has a number of limitations:
```
danielywong, line 9: Re. "mutual TLS authentication", the wording in the current [gNMI Authentication and Encryption](https://github.com/openconfig/reference/blob/master/rpc/gnmi/gnmi-authentication.md#requirements-on-the-network-device-gnmi-target) seems to imply server authentication only.

auth/authn-authz-intro.md
```diff
@@ -0,0 +1,37 @@
...
+## Proposed Direction
+
+We propose to develop an approach which addresses a number of these concerns, particularly to provide:
+
+* A secure means by which authentication and authorization policy can be provided to a network device and stored locally.
+* Means by which this policy can interact with user identifiers that are specified as part of the gRPC connection - ideally building upon wider industry developments.
+* Means by which this policy can be used as a mechanism for authenticating users against other access mechanisms (e.g., local console, SSH) such that an operator can unify their authentication and authorization policy for network devices into a single definition.
+* Means within the policy to provide for granular access to a data tree which is described by a specified data model. This will particularly focus around the YANG tree that is defined by the OpenConfig data models.
+
+We envisage that this approach can coexist with authentication and authorization mechanisms that are already implemented by typical NOS today, whilst also evolving towards a unified, modernised approach.
```
danielywong, line 27: By "coexist" I think we need to be more specific that if the new APIs are supported and in effect, control of the overlapped configurations via gNMI are required to be disabled.

auth/authn-authz-intro.md
robshakir: ACK - I think this originally was written with the context being that it was being sent to vendors. Is the clarification that we need to make here that we expect that the client also authenticates the server's CN based on its expectation. Is it CN that we're proposing to use here, or does the device itself need a cert with an associated SPIFFE ID?

auth/authn-authz-intro.md
robshakir: Thanks - I added a statement into the doc to say this. PTAL.

robshakir:
This has been replaced by `openconfig/gnsi`, closing PR.

PR #45, opened 2020-11-25
Initial discussion template for gnoi.authzn.

robshakir:
```
 * (A) auth/authzn.proto
  - Initial proposal for a local authorization and authentication
    service for network elements.
```

@aashaikh @danielywong @johnwestbrook -- PTAL.

### DIFF
--- a/auth/authzn.proto
+++ b/auth/authzn.proto
@@ -1,0 +1,179 @@
+//
+// Copyright 2020 Google Inc. All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+syntax = "proto3";
+
+import "github.com/openconfig/gnoi/types/types.proto";
+
+// This package is a work in progress. It may change in the future. :-)
+package gnoi.authzn;
+
+option (types.gnoi_version) = "0.1.0";
+
+// Service Authzn supplies the target (i.e., network device, server end of the
+// gRPC connection) with a definition of an authentication and authorization
+// policy that should be used for the respective (authn/authz) of the relevant
+// service. 
+//
+// It is expected that when authorization and authentication details are
+// supplied via the gNOI Authzn service, this config cannot be manipulated
+// through other interfaces, such that a restriction on the gNOI.authzn RPC
+// itself restricts the set of users that are able to manipulate the auth[zn]
+// configuration of the end device.
+//
+// In order to ensure that the auth[zn] configuration supplied in the service
+// cannot be subverted by changing the configuration of the device via gNMI,
+// the following leaves with the OpenConfig system model should be ignored:
+// 
+//  - /system/aaa/authentication/config/authentication-method
+//  - /system/aaa/authorization/config/authorization-method
+//  - /system/aaa/authentication/admin-user/...
+//  - /system/aaa/authentication/users/...
+//
+// The applied state (i.e., the 'state' version of the paths above) should
+// reflect a new authentication-method and authorization-method of GNOI_AUTHZN
+// which is to be added as a new identity with the base of
+// openconfig-aaa-types:AAA_METHOD_TYPE (http://ops.openconfig.net/branches/models/master/docs/openconfig-system.html#ident-aaa_method_type)
+//
+// The motivation for this service is provided in the accompanying
+// authn-authz-intro.md document.
+service Authzn {
+    // Set allows a user to specify an authentication and authorization
+    // policy to the network element. The policy, if valid, replaces the
+    // existing policy.
+    //
+    // Canonical error codes are to be used as responses. See
+    // https://grpc.io/docs/guides/error/ for more details.
+    rpc Set(SetRequest) returns (SetResponse) {}
+}
+
+message SetRequest {
+    // A list of policies that are to be applied on the network element.
+    // Multiple policies per service MAY be specified to allow for multiple
+    // users, or specific user restrictions to be described.
+    repeated Policy policy = 1;
+}
+
+// Policy specifies an authentication/authorization policy for an individual
+// service that is instantiated on the device. Some services may only have an
+// authentication policy - where they do not support more granular authorization,
+// whereas others may have both specified.
+message Policy {
+    oneof protocol {
+        // console provides the configuration for authentication of the local
+        // system login.
+        Console console = 1;
+        // ssh provides the configuration for authentication of logins to the
+        // system via SSH.
+        SSH ssh = 2;
+        // gnmi provides a local authentication mechanism for all RPCs within
+        // the gNMI service, along with an authorization policy which allows
+        // users to be restricted to:
+        //  - specific RPCs (e.g., Get, Set, Subscribe, Capabilities)
+        //  - specific restrictions based on path within each RPC.
+        gNMI gnmi = 3;
+        // gnoi provides a local authentication mechanism for each gNOI RPC
+        // within the suite of gNOI microservices whose definitions are
+        // defined in github.com/openconfig/gnoi. Within specific microservices
+        // RPC-based restrictions may be realised.
+        gNOI gnoi = 4;
+        // gribi provides a local authentication mechanism for the gRIBI service.
+        gRIBI gribi = 5;
+    }
+}
+
+message Console {
+    // Console provides a username/password based authentication mechanism
+    // for logging in to the system. 
+    //
+    // The username of the user permitted to log in to the system via the 
+    // console. If the username is specified as an
+    // empty string, the supplied authentication details should apply to the
+    // root/superuser/admin user of the system.
+    string username = 1;
+    // TODO(robjs): How should this password be supplied to the device?
+    // Presumably we do not want it in cleartext. We could re-use the same
+    // definition that is used within the OpenConfig models:
+    // http://ops.openconfig.net/branches/models/master/docs/openconfig-system.html#type-crypt-password-type
+    string password = 2;
+}
+
+message SSH {
+    // SSH provides a key-based login for users onto the system. Only users
+    // that are specified within the SSH policy are permitted to login to the
+    // system.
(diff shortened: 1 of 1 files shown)


auth/authzn.proto
```diff
@@ -0,0 +1,179 @@
...
+    // Set allows a user to specify an authentication and authorization
+    // policy to the network element. The policy, if valid, replaces the
+    // existing policy.
+    //
+    // Canonical error codes are to be used as responses. See
+    // https://grpc.io/docs/guides/error/ for more details.
+    rpc Set(SetRequest) returns (SetResponse) {}
+}
+
+message SetRequest {
```
tomek-US, line 63: How about removing a policy? Is the plan that the complete set of policies has to be provided? For all types of access?

auth/authzn.proto
```diff
@@ -0,0 +1,179 @@
...
+    //
+    // The username of a user for whom SSH login is permitted. If the value
+    // of the username field is an empty string, this corresponds to the
+    // root/superuser/admin user of the system.
+    string username = 1;
+    // authorized_keys specifies a list of SSH keys that are permitted for
+    // the user - each of which would correspond to a line within the
+    // authorized_users file on an openssh system. The CR/LF should be omitted
+    // from the supplied key.
+    repeated string authorized_keys = 2;
```
tomek-US, line 127: What about the already existing keys? Should they be removed?

auth/authzn.proto
robshakir: This would be my suggestion, yes - such that we don't have to add some form of 'Get' API to retrieve the current AAA proto, and rather just declaratively update it.

auth/authzn.proto
```diff
@@ -0,0 +1,179 @@
...
+
+message SSH {
+    // SSH provides a key-based login for users onto the system. Only users
+    // that are specified within the SSH policy are permitted to login to the
+    // system.
+    //
+    // The username of a user for whom SSH login is permitted. If the value
+    // of the username field is an empty string, this corresponds to the
+    // root/superuser/admin user of the system.
+    string username = 1;
```
tomek-US, line 122: If the user specified in the 'username' field does not exist on the box at the moment should it be created? If so, what level of access it should be given?

tomek-US: What should be done if an entry with username=="" is not present on the list? (I assume the root/admin account should not be removed)

auth/authzn.proto
```diff
@@ -0,0 +1,179 @@
...
+// The motivation for this service is provided in the accompanying
+// authn-authz-intro.md document.
+service Authzn {
+    // Set allows a user to specify an authentication and authorization
+    // policy to the network element. The policy, if valid, replaces the
+    // existing policy.
+    //
+    // Canonical error codes are to be used as responses. See
+    // https://grpc.io/docs/guides/error/ for more details.
+    rpc Set(SetRequest) returns (SetResponse) {}
```
tomek-US, line 60: Should this RPC be 'atomic' i.e. all-or-nothing-is-changed? How an error condition should be reported provided the SetResponse message is empty?

auth/authzn.proto
robshakir: For SSH access, what is the granularity of user access that you are considering? I was assuming that we are thinking that this is either allowed or not allowed - nothing more granular than this?

w.r.t to no `username=""` - I would assume that this translates to `PermitRootLogin no` in OpenSSH's `sshd_config` - rather than removing the user?

auth/authzn.proto
robshakir: Yes, I think so -- an error message could be returned using the canonical status -- which has room for a `message` [string](https://github.com/grpc/grpc/blob/master/src/proto/grpc/status/status.proto#L87)

auth/authzn.proto
```diff
@@ -0,0 +1,179 @@
...
+// policy that should be used for the respective (authn/authz) of the relevant
+// service. 
+//
+// It is expected that when authorization and authentication details are
+// supplied via the gNOI Authzn service, this config cannot be manipulated
+// through other interfaces, such that a restriction on the gNOI.authzn RPC
+// itself restricts the set of users that are able to manipulate the auth[zn]
+// configuration of the end device.
+//
+// In order to ensure that the auth[zn] configuration supplied in the service
```
danielywong, line 37: Nit on auth[zn] ... logically authentication usually comes before authorization but these words' ordering isn't consistent in this text.  Also we often use "authn" and "authz" internally but these acronyms should probably be spelt out a little more.

auth/authzn.proto
danielywong, line 60: Would it make sense to flatten the single Set API to SetSSH, SetConsole, SetXxx instead in order to allow the option to control access at the Service/RPC layer?

auth/authzn.proto
```diff
@@ -0,0 +1,179 @@
...
+        // console provides the configuration for authentication of the local
+        // system login.
+        Console console = 1;
+        // ssh provides the configuration for authentication of logins to the
+        // system via SSH.
+        SSH ssh = 2;
+        // gnmi provides a local authentication mechanism for all RPCs within
+        // the gNMI service, along with an authorization policy which allows
+        // users to be restricted to:
+        //  - specific RPCs (e.g., Get, Set, Subscribe, Capabilities)
```
danielywong, line 85: Restricting users to specific RPCs overlaps what gRPC will provide.  How do you see the two schemes work and be managed together?

auth/authzn.proto
```diff
@@ -0,0 +1,179 @@
...
+    // authorized_users file on an openssh system. The CR/LF should be omitted
+    // from the supplied key.
+    repeated string authorized_keys = 2;
+}
+
+message gRPCID {
+    // The identify of a gRPC client is defined by its SPIFFE ID (see 
+    // https://spiffe.io/docs/latest/spiffe/concepts/) which is provided within
+    // the SVID - in this case, the X.509 client certificate.
+    string id = 1;
```
danielywong, line 134: SPIFFE ID is intended to identify a workload.  As shown in https://github.com/spiffe/spiffe/blob/master/standards/SPIFFE-ID.md#22-path, some deployment may choose to encode more details in the SPIFFE ID Path components than what this authorization policy may care about.  As such, I don't think we can expect use of exact match with this "id" but allow inclusion of wildcards in the string.

auth/authzn.proto
robshakir: I was struggling with a name here that made sense to combine the two -- happy to use something different if it makes sense (`authnz` is consistent with that ordering if it's commonly used to mean this).

Once we've agreed here I'll make things consistent and spell it out - authn and authz are pretty common (used in IETF documents fairly regularly), so I don't think we have to worry too much about them being unclear.

auth/authzn.proto
robshakir: I think this would be a good reason to have them separately specified for sure -- I am not quite clear whether we expect there to be such granularity in super-user access though. Particularly, it seemed like the operational model we might expect is that there is a single group that would be responsible for all credentials for the system - like there would be today for both console/SSH local passwords and TACACS+ in most operators. If we feel that there are likely to be cases that this isn't the case, it'd be a good reason to split the RPCs.

In the scenario that we want to control access at the service layer, I assume that we actually need to allow this mechanism to define restrictions on its own contents -- such that it would be possible to allow for just SSH to be set for a particular user.

auth/authzn.proto
robshakir: My assumption here was that:

* gRPC would provide us the means to implement this authentication and authorization
* this service would provide the means to provide the policy that gRPC enforces

Is there something that explains what the proposed gRPC mechanisms are -- for example, is there already a means by which gRPC is proposing to have the per-RPC authorisation specified?

auth/authzn.proto
robshakir: (disclaimer: I'm not super-familiar with how folks are intending to use SPIFFE IDs in production)

From what you're saying - and the doc - the intention here is that we might have `spiffe://some.domain.tld/network_services/some_service/some_compute_cluster` where we want to express this policy in terms of a subset of this path (e.g.,`spiffe://some.domain.tld/network_services/some_service/*`, or even `spiffe://*.domain.tld/network_services/some_service/some_compute_cluster`). Do we solely expect that this would be wildcarding entire path components (I presume it is not going to need to be regexp based for example)?

auth/authzn.proto
danielywong: Thanks, Rob.  What I have suggested is mainly to for this specification to contain such flexibility to allow granting control to different workload/user in certain deployment.  For those who doesn't need such flexibility, the difference is basically listing 3 RPCs instead of 1 for the same rule in their gRPC authorization policy.

Sorry, I am following what you mean by "define restrictions on its own contents".

auth/authzn.proto
danielywong: Actually the gRPC authorization framework will include both policy specification and enforcement.  I will share the proposal with you separately.

Sorry for the delay in replying.  I must have have the right setup for this Github project that I am not getting email for your replies.

auth/authzn.proto
danielywong: Given spiffe-id doesn't have a very strict uri specification, I am afraid there need to be at least a way for users to specify multiple wildcards if we don't want to go as far as supporting regexp.  I think the owner identity will be the most interesting to most use cases.  Extending the "identifying service owners" example there, I can see the "ns" and "sa" kubernetes specific markers can be replaced with other such as "user-group" and "user":
spiffe://production.some_compute_cluster.example.com/group/foo/user/bar
or simply 
spiffe://production.some_compute_cluster.example.com/foo/bar

In this case, for matching all users of group Foo and regardless of which cluster the job is running from, we will want to allow users to specify:
spiffe://production.*.example.com/group/foo/user/*
or
spiffe://production.*.example.com/foo/*

What do you think?

auth/authzn.proto
tomek-US, line 60: Could we change the name of this method to, say, SetAuth() ? As it is now it prevents combining gNMI and Authz gNOI serviers into one in Go.

auth/authzn.proto
tomek-US: That still leaves question what to do if a username is not known to the device. Should the user be created? Should the whole request be rejected?

robshakir:
From our meeting on 7th January 2021:

- We will split out the RPCs so that we can have per-service authorisation on an RPC-level.
- Must users already be created before they can have their access control changed? 
  * We consider that `adduser` or another user management tool (e.g., gNMI OpenConfig system/aaa model) would give us the way to be able to add new users.
- Is the complete set to be sent, or should partial configurations be permissible?
- We need to specify both the authorised users (certificates), *and* authorised keys (public keys). We need to be able to specify both with each system user (account).

PR #49, opened 2021-04-05
Move to using protolint.

robshakir:
This PR moves to using protolint rather than the GitHub superlinter, which is overly aggressive.


robshakir:
@marcushines -- this should be ready for review now. Fixed linter, and moved to modern `protoc`/`protoc-gen-go`/`protoc-gen-grpc-go`.

.github/workflows/linter.yml
```diff
@@ -18,9 +18,9 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Lint Code Base
-        uses: github/super-linter@v3
+      - name: Lint protobuf
+        uses: robshakir/protolint-action@v0.3.1
```
marcushines, line 22: why not use something not you for hosting?

marcushines (approved):

.github/workflows/linter.yml
robshakir: We can create an `openconfig` repo for this, but I'm going to submit a PR to the original owner of this action so that I can remove the fork -- it just looks like it hasn't kept pace with the releases of protolint.

marcushines: sgtm

robshakir: PR submitted upstream and now back to using the upstream action.

PR #51, opened 2021-04-13
Added the 'no_reboot` flag to the ActivateRequest message.

tomek-US:
If this flag is set to 'False' then the Target will configure itself
to boot next time with the requested version of the OS but will _not_
restart immediatelly.


os/os.proto
```diff
@@ -250,11 +251,18 @@ message InstallError {
...
   string version = 1;
   // For dual Supervisors setting this flag instructs the Target to perform the
   // action on the Standby Supervisor.
   bool standby_supervisor = 2;
+  // If set to 'True' the Target will initiate the reboot process immediatelly
+  // after changing the next bootable OS version.
+  // If set to 'False' a separate action to reboot the Target and start using
+  // the activated OS version is required. This action CAN be executing
+  // the gNOI.system.Reboot() RPC.
+  bool activate_and_reboot = 3;
```
samribeiro, line 265: The inverse logic should be used here in order to keep the same behaviour with backward compatibility. Something like "no_reboot" or "activate_only" (proto3 defaults boolean to false when omitted). Where by default the Activate request will activate and reboot the target, unless this flag is set in which only activation will happen, but no reboot. This allows previous and actual versioned client and target implementations to coexist without a fundamental behaviour change of the RPC like it is being proposed here.

os/os.proto
```diff
@@ -7,7 +7,7 @@ import "github.com/openconfig/gnoi/types/types.proto";
 
 option go_package = "github.com/openconfig/gnoi/os";
 
-option (types.gnoi_version) = "0.1.0";
+option (types.gnoi_version) = "0.2.0";
```
samribeiro, line 10: assuming the suggestion below, I am thinking this would be a micro bump and opposed to minor.

os/os.proto
tomek-US: Keeping backward compatibility sounds like good idea ;-) Done.

os/os.proto
tomek-US: Done

samribeiro (approved):
thanks!

samribeiro:
Probably worth regenerating the go code for this proto.

tomek-US:
Thx!

The second commit already includes re-generated Go files.

Issue #60, opened 2022-01-14
Inconsistent specification of Ping count of 0


greg-dennis:
In the doc of the Ping RPC is says "If a packet count is not explicitly provided, 5 is used":
https://github.com/openconfig/gnoi/blob/master/system/system.proto#L39

In the doc of PingRequest it says "A count of 0 defaults to a vendor specified value, typically 5."
https://github.com/openconfig/gnoi/blob/master/system/system.proto#L172

So must a count of zero use 5 or is it really left up to the vendor?

Regardless, I recommend removing any mention of the count from the RPC itself, so that the code only contains one source of truth for what a zero count means.

marcushines:
Going to remove the line in the rpc spec

PR #61, opened 2022-01-14
Update system proto to address

marcushines:
https://github.com/openconfig/gnoi/issues/60

this removes the confusing comment in the rpc specification for Ping


robshakir (approved):
LGTM, thanks.

PR #63, opened 2022-01-31
Update workflow to use 1.17

marcushines:


.github/workflows/go.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: ['1.12', '1.13', '1.14', '1.15']
+        go: ["1.16","1.17"]
```
robshakir, line 15: You can just write `1.x` here if you want to run for the latest version.

robshakir (approved):

Issue #64, opened 2022-03-08
healthz.proto response should be streaming type


amanand:
In order to export core files or large outputs like "show tech-support" as part of the RPC response, shouldn't the rpc be server streaming type?
https://github.com/openconfig/gnoi/blob/7dd3a95a4f1ee589048ac8ebe6ac84c154c152df/healthz/healthz.proto#L43

marcushines:
Addressed in the streaming healthz PR https://github.com/openconfig/gnoi/pull/79

PR #65, opened 2022-03-19
Update healthz `Get` RPC to streaming response API

earies:
As mentioned in the issue: https://github.com/openconfig/gnoi/issues/64

This proposal is to initiate discussion on appropriate design for this API since the intention is that implementor extensions will bring in the capability to return large data sets or full blown file transfer thus duplicating behaviors seen in the gnoi.file services.

If the intention is to keep these capabilities in these services, then it is worth exploring duplication of how the `Get` RPC is defined in the gnoi.file `File` service.

Note: This PR does not currently contain generated stubs or other carried over characteristics from the gnoi.file `File` service

### DIFF
--- a/healthz/healthz.proto
+++ b/healthz/healthz.proto
@@ -30,17 +30,17 @@ import "google/protobuf/any.proto";
 
 option go_package = "github.com/openconfig/gnoi/healthz";
 
-option (types.gnoi_version) = "1.0.0";
+option (types.gnoi_version) = "2.0.0";
 
 // The Healthz service provides access to the status of a path on the
-// system. Addtitionally it allows the implementor to provide path specific
-// diagnositic data into the status return.
+// system. Additionally it allows the implementor to provide path specific
+// diagnostic data into the status return.
 //
 // Healthz is expected to work in conjunction with the component OC model.
 service Healthz {
   // Get will get health status for a gNMI path.  If no status is available for
   // the requested path an error will be returned.
-  rpc Get(GetRequest) returns (GetResponse) {}
+  rpc Get(GetRequest) returns (stream GetResponse);
 }
 
 message GetRequest {


earies:
@marcushines - can you PTAL - thx

tomek-US:
The idea of having a streaming version of the `Get` operations looks good to me, however, to keep backward compatibility, it should be an _additional_ RPC.

marcushines:
I would worry about making "yet another streaming rpc protocol"

Do you have a concrete example of what exact data we are talking about here?

tomek-US:
I think the problem described in #64 is a valid concern - there are limits on how much of data can be returned in one proto - that's a reason why we have `gNMI.Subscribe(ONCE)` after all, but the `GetResponse` message should be modified to allow sending partial responses.

earies:
> The idea of having a streaming version of the `Get` operations looks good to me, however, to keep backward compatibility, it should be an _additional_ RPC.

Agreed as this PR stands would be a backwards incompatible change however I would then solicit how much adoption these APIs have gotten in reality if preserving the old API and introducing another is actually warranted

> I would worry about making "yet another streaming rpc protocol"
> Do you have a concrete example of what exact data we are talking about here?

Precisely my concern as well.  Within the `ComponentStatus` message we currently have the following field:

```
  // Opaque data for how the healthcheck is implemented.  This can be any proto
  // defined by the vendor.  This could be the equivalent to outputs like
  // "show tech" or core files or any other diagnostic data.
  google.protobuf.Any healthz = 4;
```

So in the case of proprietary debug information as mentioned above, the implementor has the opportunity to do what they wish here which could result in a large amount of data transfer including but not limited to the actual transfer of files (e.g. core files) on the filesystem.

Another option could be to *not* create another method for file transfer and leverage the existing `File` services where the `ComponentStatus` could indicate via hints either directly as a well-defined message or vendor extension of the absolute path(s) of interest to then call in a subsequent `File.Get()`

So far the `Healthz` service mostly defines a method to `Get` targeted specifically at the `/components` subtree where all defined return messages would be the same as what one could `Get` over gNMI to that same subtree targeted specifically for YANG nodes as being worked in https://github.com/openconfig/public/pull/504 (gNMI also brings the advantage that these nodes would be `ON_CHANGE` capable).  So as I see it, we have some duplication with the exception of the `Any` extension that may also be able to be extended by way of YANG augments or similar concepts as mentioned above (file system hints).

Is a gNOI service needed at all then for this purpose or can gNMI solely be leveraged (possibly in conjunction with the `File` gNOI service)?

earies:
@marcushines @tomek-US - curious to any comments on above.  If we don't factor in any proprietary protobuf extensions here, the service as it stands is no different than a gNMI unary `Get` to the same `/components` path supporting any modeled health parameters which was the intention of https://github.com/openconfig/public/pull/504 (albeit some current misalignment)

tomek-US:
Why are you worried about "yet another  streaming rpc protocol"?

I agree that at some angle there is no difference between those two approaches (gNMI vs gNOI) but as this information might be sensitive it seems to be that it will be easier to control the access to this data when gNOI is used.

Moreover, the usage of the dedicated API for getting the diagnostics data eliminates a need to define location/naming of the files with this data if a generic `File` API is used.

marcushines:
so your point on auth control is actually a reasonable point on why this might be worth developing as a good use case.
so the interface though for the protocol would need to significantly change to fully support a streamed response properly

earies:
My concern is less on "yet another streaming protocol" but rather the intention and overlap to be articulated and aligned with the YANG model work in the other referenced PR and the value this service brings in general above what can be accomplished in gNMI.  This PR is infact initiated to propose turning this into a streaming RPC which appears an oversight on initial design given the comments and intention for the extension field (to transfer what can be very large amounts of data)

This service definition as I see it is only of use if the extension data is leveraged and thus this turns into a streaming RPC but it also seems that specific protocol needs to be put in place regarding file transfer (block size, hashes, etc..) much like what is defined in the `File` service thus questioning if we need to redefine these same behaviors here

marcushines:
This PR is being addressed via 

https://docs.google.com/document/d/1xivx0SWuNrBcPuGmQP7M2XSlyve5xWrsckgRWwCTUiM/edit#heading=h.x9snb54sjlu9

PR #66, opened 2022-04-12
Add network instance to ping and traceroute

rsgcp:


system/system.proto
```diff
@@ -184,6 +184,7 @@ message PingRequest {
   bool do_not_fragment = 7; // Set the do not fragment bit. (IPv4 destinations)
   bool do_not_resolve = 8;  // Do not try resolve the address returned.
   types.L3Protocol l3protocol = 9; // Layer3 protocol requested for the ping.
+  string networkInstance = 10; //  vrf
```
earies, line 187: Would suggest common nomenclature for this field across all gnxi APIs and also one that aligns w/ the schema set.  Currently `bgp.proto` appears to be the only gNOI API that has a concept of `network-instance` however calls this `routing_instance`.

I'd suggest fixing `bgp.proto` and keeping the nomenclature here and there to a field of `network_instance` for consistency using snake_case

earies:
The title suggests this is for traceroute as well so it looks like `TracerouteRequest` should be modified in similar fashion

PR #68, opened 2022-04-22
add packet_link_qualification

yushiyushi:
Copy packet_link_qualification to gnoi


qualification/packet_link_qualification.proto
```diff
@@ -0,0 +1,209 @@
+// A generic network operational interface gRPC service to perform packet based
+// link qualification operations on a network device.
+
+syntax = "proto3";
+
+package gpins.gnoi;
```
romeyod, line 6: This should be "package gnoi.qualification;"

qualification/packet_link_qualification.proto
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+  message QualificationConfiguration {
+    // The ID assigned to the given test. The ID value must be unique among the
+    // tests running on this agent.
+    string id = 1;
+
+    // Path to the interface corresponding to the port to test.
+    .gnoi.types.Path interface = 2;
+
```
romeyod, line 89: Are we planning to add "packet_size" in different patch?

rsgcp: We can add it to this version. I will work with Yushi to add it here.

romeyod:
We will also need changes to "[gnoi](https://github.com/openconfig/gnoi)/compile_protos.sh"

This brings up another issue: the file naming convention so far has been directory name is the same as the proto file name for other gnoi. Example: gnoi/factory_reset/factory_reset.proto
We need to make sure the contributors/maintainers are okay with this difference of qualification/packet_link_qualification.proto

packet_link_qualification/packet_link_qualification.proto
```diff
@@ -0,0 +1,212 @@
...
+    string id = 1;
+    // Status of the operation. In case of failures, canonical error code and
+    // message will show the details.
+    google.rpc.Status status = 2;
+    // Qualification results.
+    QualificationResults results = 3;
+  }
+
+  repeated GetResponse responses = 1;
+}
```
marcushines, line 212: add blank new line

marcushines (approved):

PR #69, opened 2022-04-27
Add do_not_lookup_asn flag to TracerouteRequest

brianneville:
The do_not_lookup_asn bool can be set to true in the
 TracerouteRequest to indicate that the peer should not
 attempt to lookup ASNs during the traceroute.


dplore (approved):
LGTM @marcushines for any additional comment

PR #73, opened 2022-05-02
Proposed updates to LinkQual proto

robshakir:
```
commit cb737eb4956b2e96f17a803bf76ffb2bbe74fb08
Author: Rob Shakir <robjs@google.com>
Date:   Mon May 2 16:52:24 2022 -0700

    Best practice naming and typing.
    
     * (M) packet_link_qualification/packet_link_qualification.proto
      - Use singular naming for repeated fields.
      - switch uint16 (no such time) to uint32.

commit 49325ffe5408ec27a0302612b2530a21ce439537
Author: Rob Shakir <robjs@google.com>
Date:   Mon May 2 16:50:04 2022 -0700

    Proposed updates for packet_link_qualification.
    
     * (M) packet_link_qualification/packet_link_qualification.proto
      - Do not embed google.rpc.Status as a payload for an error, and rather just
        use the status that is returned by gRPC. This avoids the issue of having
        a "successful" RPC that actually failed and matches other g* protocols.
      - Remove nested messaging.
      - Adopt best practice naming for repeated and enums.
```


packet_link_qualification/packet_link_qualification.proto
```diff
@@ -70,98 +78,92 @@ enum QualificationState {
...
+  // tests running on this agent.
+  string id = 1;
+
+  // Path to the interface corresponding to the port to test.
+  types.Path interface = 2;
+
+  // The minimum amount of wait time before sending packets on the link. It
+  // gives peer additional time to prepare. The default value for this field
+  // is 70 seconds.
+  int32 minimum_wait_before_preparation_seconds = 3;
```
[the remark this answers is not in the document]
marcushines: after thinking on this more if our expectation is that the device after start may become unreachable should we break this into a setup step then a start step this would let you setup both sides before starting both sides
where the setup could validate that device capable of doing the work vs. the start setting up and "waiting to start" then the other side try to setup and fails then you have "quickly" jump back to the peer. device and stop

packet_link_qualification/packet_link_qualification.proto
```diff
@@ -70,98 +78,92 @@ enum QualificationState {
...
-  repeated StartResponse responses = 1;
+  repeated StartResponse response = 1;
+}
+
+message StartResponse {
+  // The same ID given in the start request.
+  string id = 1;
 }
 
 message StopPacketQualificationRequest {
```
marcushines, line 149: should this be cancel? vs. stop?

marcushines: what it is noted above that stop clears the stats - that feels conterinitive from a workflow

seems like it would be

Setup
Start
Get (to see if things are progressing) (may not be available)
Stop (may not be available)
Get (once the system reachability has been guaranteed to be restored
Clear

marcushines:
Closing as v2 has been submitted

PR #79, opened 2022-05-24
Update healthz based on design doc

marcushines:
update index.md to match

### DIFF
--- a/healthz/healthz.proto
+++ b/healthz/healthz.proto
@@ -41,6 +41,20 @@ service Healthz {
   // Get will get health status for a gNMI path.  If no status is available for
   // the requested path an error will be returned.
   rpc Get(GetRequest) returns (GetResponse) {}
+
+  // List returns all events for the provided component.
+  rpc List(ListRequest) returns (ListResponse) {}
+
+  // Acknowledge will set the acknowledged field for the event.
+  // This is an idempotent operation.
+  rpc Acknowledge(AcknowlegeRequest) returns (AcknowledgeResponse) {}
+
+  // Artifact will stream the artifact contents for the provided artifact id.
+  rpc Artifact(ArtifactRequest) returns (stream ArtifactResponse) {}
+
+  // Check will invoke the healthz on the provided component. This RPC can be
+  // expensive depending on the vendor implementation.
+  rpc Check(CheckRequest) returns (CheckResponse) {}
 }
 
 message GetRequest {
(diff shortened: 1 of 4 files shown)


healthz/healthz.proto
```diff
@@ -50,24 +65,134 @@ message GetRequest {
...
+  // Local file name of the artifact.
+  string name = 1;
+  // Path to file on the local file system. (optional)
+  string path = 2;
+  // Mimetype of the file.
+  string mimetype = 3;
+  // Size of the file.
+  int64 size = 4;
+  // Hash of the file.
+  types.HashType hash = 5;
```
brianneville, line 175: Should we specify what hash function to use here?
If the client and server don't know which hash function to use, then we cannot validate the hash on the client-side

healthz/healthz.proto
```diff
@@ -50,24 +65,134 @@ message GetRequest {
...
+  // The individual artifacts can be retrieved via the Artifact() RPC.
+  repeated ArtifactHeader artifacts = 5;
+
+  // ID is the unique key for this event in the system.
+  string id = 6; 
+
+  // Acknowledged is set when a caller has processed the event.
+  bool acknowledged = 7;
+
+  // Create is the timestamp when this event was created.
```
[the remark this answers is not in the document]
marcushines: fixed

healthz/healthz.proto
marcushines: that ^^^

healthz/healthz.proto
```diff
@@ -41,6 +42,20 @@ service Healthz {
   // Get will get health status for a gNMI path.  If no status is available for
   // the requested path an error will be returned.
   rpc Get(GetRequest) returns (GetResponse) {}
+
+  // List returns all events for the provided component.
```
robshakir, line 46: Should this be `provided path`, to be consistent with the line above?

healthz/healthz.proto
```diff
@@ -41,6 +42,20 @@ service Healthz {
...
+  rpc List(ListRequest) returns (ListResponse) {}
+
+  // Acknowledge will set the acknowledged field for the event.
+  // This is an idempotent operation.
+  rpc Acknowledge(AcknowlegeRequest) returns (AcknowledgeResponse) {}
+
+  // Artifact will stream the artifact contents for the provided artifact id.
+  rpc Artifact(ArtifactRequest) returns (stream ArtifactResponse) {}
+
+  // Check will invoke the healthz on the provided component. This RPC can be
```
robshakir, line 56: `provided path` if we decide this is the right nomenclature.

healthz/healthz.proto
```diff
@@ -50,24 +65,134 @@ message GetRequest {
...
+  google.protobuf.Any healthz = 4 [deprecated=true];
+
+  // Artifacts provides links to all artifacts contained in this event.
+  // The individual artifacts can be retrieved via the Artifact() RPC.
+  repeated ArtifactHeader artifacts = 5;
+
+  // ID is the unique key for this event in the system.
+  string id = 6; 
+
+  // Acknowledged is set when a caller has processed the event.
```
robshakir, line 114:
```suggestion
  // Acknowledged is set when at least one caller has processed the event.
```

healthz/healthz.proto
```diff
@@ -50,24 +65,134 @@ message GetRequest {
...
+  bool acknowledged = 7;
+
+  // Created is the timestamp when this event was created.
+  google.protobuf.Timestamp created = 8;
+  
+  // Expires is the timestamp when the system will clean up the
+  // artifact. If unset, the artifact is not scheduled for GC.
```
robshakir, line 121:
```suggestion
  // artifact. If unset, the artifact is not scheduled for garbage
  // collection.
```

healthz/healthz.proto
```diff
@@ -50,24 +65,134 @@ message GetRequest {
...
+
+message ArtifactResponse {
+  oneof contents {
+    // Header is the first message in the stream. It contains
+    // the id of the artifact and metadata for the artifact
+    // based on the type of the artifact.
+    // OC defines FileArtifactType and ProtoArtifactType.
+    ArtifactHeader header = 1;
+    // Trailer contains the checksum for the streamed artifact.
+    Trailer trailer = 2;
```
robshakir, line 159: +1 to Brian's comment above, but also, should this be `ArtifactTrailer` if we keep it to ensure that this is consistently named?

healthz/healthz.proto
```diff
@@ -50,24 +65,134 @@ message GetRequest {
...
+}
+
+// Generic proto message artifact stream.
+// This proto tells the caller that the artifact stream 
+// will be a stream of proto encoded messages that make up
+// the artifact. Each message must be deserialized by the caller
+// and there are no other assumptions about the number of
+// messages or length of the stream or how those messages are to
+// be reassembled.
+message ProtoArtifactType {
```
robshakir, line 185: What was the motivation to have this rather than just use the `proto.Any`? Since this doesn't carry any additional information, do we expect that this is distinguishable at all from just having `bytes`?

robshakir (changes requested):

healthz/healthz.proto
```diff
@@ -50,24 +65,134 @@ message GetRequest {
...
+}
+
+message ArtifactResponse {
+  oneof contents {
+    // Header is the first message in the stream. It contains
+    // the id of the artifact and metadata for the artifact
+    // based on the type of the artifact.
+    // OC defines FileArtifactType and ProtoArtifactType.
+    ArtifactHeader header = 1;
+    // Trailer contains the checksum for the streamed artifact.
```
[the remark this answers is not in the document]
marcushines: it is not there this was just left over comment

healthz/healthz.proto
```diff
@@ -41,6 +41,20 @@ service Healthz {
   // Get will get health status for a gNMI path.  If no status is available for
```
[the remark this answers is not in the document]
marcushines: 2.0.0 would be on removal of the deprecated field
this doesn't break any capability i would say 1.1

healthz/healthz.proto
marcushines: done

healthz/healthz.proto
marcushines: done

healthz/healthz.proto
marcushines: this is the header definition so a protoartifacttype has a clear definition on what it means
custom is custom in that it can define it's own encoding scheme past the defined proto schema which says
you must provide only a stream of proto messages 

vs. custom you could define whatever y our want

robshakir: Thanks for explaining, LGTM.

robshakir (approved):

PR #85, opened 2022-06-29
Clarify ping RPC expected behavior

bstoll:
While looking into a test to verify ping behavior, I became unsure of what is expected in the PingResponse stream.  It looks like it would be valid to return individual packets OR summary statistics OR both the individual packets and the summary.

This change is to clarify that summary statistics are always expected, and individual packet metrics should be returned (ie, it's probably not reasonable for a flood request).

### DIFF
--- a/system/system.proto
+++ b/system/system.proto
@@ -36,7 +36,8 @@ option (types.gnoi_version) = "1.0.0";
 service System {
   // Ping executes the ping command on the target and streams back
   // the results.  Some targets may not stream any results until all
-  // results are in.
+  // results are in.  The stream should provide single ping packet responses
+  // and must provide summary statistics.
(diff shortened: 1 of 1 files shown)


hellt:
Since there is no option in the `PingRequest` to ask for a flood ping, shall we assume that we also always request single ping stats to be returned? 
The way I read the spec is that both single ping responses and summary responses are expected

bstoll:
A flood ping is specified by [setting "interval" to -1](https://github.com/openconfig/gnoi/blob/18d7d51539452c1417ef469877f0f3929f09223f/system/system.proto#L174).

hellt:
my bad @bstoll 
I can't read

marcushines (approved):

PR #94, opened 2022-09-08
gnoi.Healthz: fix typo `healthly->healthy`

hellt:
Hi @marcushines 
It looks like this is a typo.

**Note:** I didn't regen the protos, as I find the compile protos script a bit unfriendly to contributors:

- no desired protoc version set
- no instructions to setup deps

Ideally, I think it makes a lot of sense to use a protoc gen in a container, so that everyone can use the same hermetic version of protoc to generate protos. 

If my preference match yours I can contribute this piece to gnoi


marcushines:
I think that is reasonable - another suggestion would be to switch to bazel for this as well

hellt:
@marcushines would someone from OC with the protoc toolchain set generate protos to complete this PR while docker/bazel is in progress?

marcushines (approved):

PR #99, opened 2022-09-30
Update Reboot proto

xw-g:
Propose to:
* Clarify the behavior when the list of subcomponents is not set.
* Because `RebootStatusRequest` supports a list of subcomponents, the `RebootStatusResponse` should also have a way to identify status per subcomponent.

Another option is to deprecate the list of subcomponents in all reboot related proto, and only allow the request to target a single subcomponent. But then it raises more questions, like:
* Can the single subcomponent be an optional field?
* If it can be an optional field, what the default behavior if it's not set? 
* If the default behavior is to apply to the whole device (including all subcomponents), then any status reports should also be per component anyway.

### DIFF
--- a/system/system.proto
+++ b/system/system.proto
@@ -108,7 +108,8 @@ message RebootRequest {
   uint64 delay = 2;
   // Informational reason for the reboot.
   string message = 3;
-  // Optional sub-components to reboot.
+  // Optional, sub-components to reboot.
+  // If it's not set, reboot the whole network device.
   repeated types.Path subcomponents = 4;
   // Force reboot if sanity checks fail. (ex. uncommited configuration)
   bool force = 5;
@@ -138,7 +139,9 @@ enum RebootMethod {
 // request.
 message CancelRebootRequest {
   string message = 1;      // informational reason for the cancel
-  repeated types.Path subcomponents = 2; // optional sub-components.
+  // Optional, sub-components to cancel reboot.
+  // If it's not set, cancel all outstanding reboot requests.
+  repeated types.Path subcomponents = 2;
 }
 
 message CancelRebootResponse {
@@ -149,11 +152,21 @@ message RebootStatusRequest {
 }
 
 message RebootStatusResponse {
-  bool active = 1;      // If reboot is active.
-  uint64 wait = 2;      // Time left until reboot.
-  uint64 when = 3;      // Time to reboot in nanoseconds since the epoch.
-  string reason = 4;    // Reason for reboot.
-  uint32 count = 5;     // Number of reboots since active.
+  bool active = 1 [deprecated = true];  // If reboot is active.
+  uint64 wait = 2 [deprecated = true];  // Time left until reboot.
+  uint64 when = 3
+      [deprecated = true];  // Time to reboot in nanoseconds since the epoch.
+  string reason = 4 [deprecated = true];  // Reason for reboot.
+  uint32 count = 5 [deprecated = true];   // Number of reboots since active.
+
+  message Status {
+    bool active = 1;    // If reboot is active.
+    uint64 wait = 2;    // Time left until reboot.
+    uint64 when = 3;    // Time to reboot in nanoseconds since the epoch.
+    string reason = 4;  // Reason for reboot.
+    uint32 count = 5;   // Number of reboots since active.
+  }
+  repeated Status statuses = 6;
 }
 
 // A TimeRequest requests the current time accodring to the target.


xw-g:
@marcushines

xw-g:
This will probably impact the following merged PRs:
* https://github.com/openconfig/featureprofiles/pull/281
* https://github.com/openconfig/featureprofiles/pull/124

system/system.proto
```diff
@@ -149,11 +152,21 @@ message RebootStatusRequest {
...
+  uint32 count = 5 [deprecated = true];   // Number of reboots since active.
+
+  message Status {
+    bool active = 1;    // If reboot is active.
+    uint64 wait = 2;    // Time left until reboot.
+    uint64 when = 3;    // Time to reboot in nanoseconds since the epoch.
+    string reason = 4;  // Reason for reboot.
+    uint32 count = 5;   // Number of reboots since active.
+  }
+  repeated Status statuses = 6;
```
marcushines, line 169: #not sure the logics correctly - if send a statusrequest that is empty what is the expected return? a statuses with a len of 1? 

what if there is no reboot in process?
does the system return a single status message saying no reboot or empty statuses

marcushines: should the Status have the path that is been queried as part of the response?

system/system.proto
```diff
@@ -108,7 +108,8 @@ message RebootRequest {
   uint64 delay = 2;
   // Informational reason for the reboot.
   string message = 3;
-  // Optional sub-components to reboot.
+  // Optional, sub-components to reboot.
+  // If it's not set, reboot the whole network device.
```
marcushines, line 112: hmm i feel like we should require the component to reboot since rebooting a chassis as the default seems like a rather large hammer to "default" to

xw-g: This PR does require the user/client to be careful with the big hammer. 

I had the similar concern, but I also debated with myself that if the field is required, then we probably want to have a very standard way (across all the vendors) to say this `Foo` component/sub-component means the whole device for all vendors? Is it safe to use ` /components/component/name (identity=CHASSIS)` here? @dplore

system/system.proto
xw-g: * ah, good catch. I copied and pasted the fields too fast and forgot to add the subcomponent paths. I just added it. 

* Reading the existing fields (e.g. `count, active`), seems like the original idea is to return reboot status (i.e. active=false) even no active reboot or planned reboot on the sub-component? The original idea makes sense to me. `RebootStatusRequest` reports all reboots activities which covers the history and planning. Therefore, to align with the original idea I think we can clarify here saying that if sub-component not set, return status on all sub-components. Thoughts?

AnandJyrm: @xw-g The last entry of Reboot Status is covered in the openconfig-platform: last-reboot-reason.

```proto
    leaf last-reboot-reason {
      type identityref {
        base oc-platform-types:COMPONENT_REBOOT_REASON;
      }
      description
        "This reports the reason of the last reboot of the component.";
    }
```
Should the RebootStatusRequest be applicable to only current/planned reboot requests?

system/system.proto
hellt: If we use some vendor into consideration, the chassis might have the following component - CPM that is equipped in a given slot (in a redundant CPM mode). Which is modelled as `/platform/control[slot=*]`.
So CPM_A would be `/platform/control[slot=A]` and CPM_B would be at `/platform/control[slot=B]`

1. What would be the component path in OC notation? Will the individual CPMs in their respective slots be modelled as subcomponents or individual components?
2. in any case, if we put a say `component[name=control-plane-module]` in a request and do not specify any potential subcomponents, then rebooting the whole network device seems like an overkill. Should it rather be a reboot of all subcomponents instead?

system/system.proto
```diff
@@ -149,11 +152,24 @@ message RebootStatusRequest {
...
+  bool active = 1 [deprecated = true];  // If reboot is active.
+  uint64 wait = 2 [deprecated = true];  // Time left until reboot.
+  uint64 when = 3
+      [deprecated = true];  // Time to reboot in nanoseconds since the epoch.
+  string reason = 4 [deprecated = true];  // Reason for reboot.
+  uint32 count = 5 [deprecated = true];   // Number of reboots since active.
+
+  message Status {
+    types.Path subcomponent =
+        1;              // sub-component that the reboot status is for.
```
hellt, line 164: Unnecessary newline?

system/system.proto
hellt: Should the `subcomponent` be renamed to `component`? What is the logic behind naming it a `subcomponent`? In openconfig-platform we have `component -> subcomponent` relationship. If we assume that reboot works on paths defined in openconfig-platform, then maybe using a subcomponent name is not entirely logical.

system/system.proto
marcushines: i would return error here if unset - as the total number of components may be large

system/system.proto
marcushines: yes component makes sense - and if not set return not found or invalid parameter

hellt: then I suppose to reboot the chassis users should use the `/components/component[name=chassis]` or something similar to indicate the the control plane modules should reboot?

marcushines: yes

hellt:
Hi all,
Since this PR is targeting the Reboot RPC, I would like to clarify what is the common thinking for the RebootResponse when you reboot the active CPM or the whole chassis.

If gNOI server runs on CPM_A, then what is expected of a system when you reboot this CPM or the whole chassis? Should the response be sent back before the component goes for a reboot? or should it just go for a reboot which results in a TCP session to break?

/cc @marcushines

system/system.proto
dplore: We should require a single component.  If the component is not specified, there should be an error.  

Also agreed that we should explicitly clarify that reboot of /components/component[name=chassis]  indicates a full system reboot.

dplore, line 112:
```suggestion
  // This string must contain the /components/component/name to reboot
  // If not set, INVALID_ARGUMENT should be returned.
  // If the request is accepted by the device, a gRPC OK should be returned.
  // Reboot of /components/component[name=chassis] indicates a full system reboot.
  // In the case of a full device reboot, a gRPC OK should still be returned and the gRPC 
  // session closed.  This indicates to the client that the request was accepted and the 
  // reboot will proceed.
  // 
```

system/system.proto
```diff
@@ -149,11 +152,24 @@ message RebootStatusRequest {
...
-  uint32 count = 5;     // Number of reboots since active.
+  bool active = 1 [deprecated = true];  // If reboot is active.
+  uint64 wait = 2 [deprecated = true];  // Time left until reboot.
+  uint64 when = 3
+      [deprecated = true];  // Time to reboot in nanoseconds since the epoch.
+  string reason = 4 [deprecated = true];  // Reason for reboot.
+  uint32 count = 5 [deprecated = true];   // Number of reboots since active.
+
+  message Status {
+    types.Path subcomponent =
```
dplore, line 163: Can we use component instead of sub-component?  A flat list of components should do the job.  Any use cases for hierarchy of components can be determined via gNMI.

dplore: Component should be mandatory.  The component tree is very large on most devices.

dplore (changes requested):

system/system.proto
AnandJyrm: @dplore 
```
  // In the case of a full device reboot without delay, a gRPC OK should still be returned and the gRPC 
  // session closed.  This indicates to the client that the request was accepted and the 
  // reboot will proceed. With delay, gRPC OK will be returned
  ```

xw-g:
Internal bug b/245550570.

LimeHat:
@dplore I'm curious, is there a plan to move this PR forward? Been stale for almost a year.

PR #105, opened 2022-10-26
Add support for SIGABRT to KillProcessRequest_Signal, allow gNOI to generate coredumps

aredmon8551:
All currently defined signals only have an action of terminate, none generate a core.   Proposing addition of SIGABRT to support core generation:
https://man7.org/linux/man-pages/man7/signal.7.html


system/system.proto
```diff
@@ -329,6 +329,7 @@ message KillProcessRequest {
     SIGNAL_TERM = 1;         // Terminate the process gracefully.
     SIGNAL_KILL = 2;         // Terminate the process immediately.
     SIGNAL_HUP = 3;          // Reload the process configuration.
+    SIGNAL_ABRT = 4;         // Terminate the process immediately, generate core.
```
hellt, line 332: Maybe clarify the core concept a little bit more?
```suggestion
    SIGNAL_ABRT = 4;         // Terminate the process immediately, and generate a core dump file.
```

marcushines (approved):

hellt:
@marcushines  @aredmon8551  We also need bump the service version (and regen the protos?)

marcushines:
Please fix lint as well as generate protos 

I don't know that you have to bump as this is not breaking

hellt:
@marcushines shouldn't new functionality (like supporting a new signal) require a bump in the minor version of a semver-styled version string?

I.e. in our product documentation we aim to provide a service version which we support, i.e. `1.2.0` plus the commit hash which we used to get the protos from.
Now if a new signal is added while the version stays the same `1.2.0`, someone may expect to use that signal, judging by the version match.

system/system.pb.go
```diff
@@ -245,6 +245,7 @@ const (
 	KillProcessRequest_SIGNAL_TERM        KillProcessRequest_Signal = 1 // Terminate the process gracefully.
 	KillProcessRequest_SIGNAL_KILL        KillProcessRequest_Signal = 2 // Terminate the process immediately.
 	KillProcessRequest_SIGNAL_HUP         KillProcessRequest_Signal = 3 // Reload the process configuration.
+	KillProcessRequest_SIGNAL_ABRT        KillProcessRequest_Signal = 4 // Terminate the process immediately and dump a core file.
```
hellt, line 248:
```suggestion
	KillProcessRequest_SIGNAL_ABRT        KillProcessRequest_Signal = 4 // Terminate the process immediately and dump a core file.
```

system/system.pb.go
```diff
@@ -254,12 +255,14 @@ var (
...
 		2: "SIGNAL_KILL",
 		3: "SIGNAL_HUP",
+                4: "SIGNAL_ABRT",
 	}
 	KillProcessRequest_Signal_value = map[string]int32{
 		"SIGNAL_UNSPECIFIED": 0,
 		"SIGNAL_TERM":        1,
 		"SIGNAL_KILL":        2,
 		"SIGNAL_HUP":         3,
+                "SIGNAL_ABRT":        4,
```
hellt, line 265: tabs instead of spaces
```suggestion
		4: "SIGNAL_ABRT",
	}
	KillProcessRequest_Signal_value = map[string]int32{
		"SIGNAL_UNSPECIFIED": 0,
		"SIGNAL_TERM":        1,
		"SIGNAL_KILL":        2,
		"SIGNAL_HUP":         3,
		"SIGNAL_ABRT":        4,
```

hellt (commented):
@marcushines here is a suggestion set for fixing gofmt complaints. Although the question remains what protoc-gen code made it non-compliant? Looks weird

system/system.pb.go
aredmon8551: Fixed

system/system.pb.go
aredmon8551: Fixed

aredmon8551:
Regenerated protos, but believe some things are wrong here:
1. The go.mod file has been updated (https://github.com/openconfig/gnoi/pull/105/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R7) to include new dependencies.  This doesn't seem right - any insight into how to prevent this would be appreciated.
2. I had to manually pull down the referenced projects in the compile.sh script to my $GOPATH because `go get` fails (as I believe no main.go file exists in those projects).  Am I doing this right?

PR #106, opened 2022-11-01
Update healthz documentation.

robshakir:
```
* (D) healthz/index.md
  - Merge with README.md
 * (M) healthz/README.md
  - Update README to cover latest changes to healthz proto.
  - Add additional documentation describing RPC implementations.
 * (M) healthz/healthz.proto
  - Editorial updates.
```

@hellt @marcushines -- PTAL. This updates the documentation with the decisions made in the implementation of streaming healthz.


healthz/README.md
```diff
@@ -7,41 +14,239 @@ in distributed systems.
...
+on components and if "unhealthy" gather implementor specific data to help
+triage or reproduce issues.
+
+## Design Background
+
+`Healthz` provides a means by which a user may initiate checks of health
+(through the `Check` RPC) or a system may report the results of a check that it
+has initiated of its own accord. 
+
+Following a health check occuring, a caller uses the `List` RPC to discover the
```
hellt, line 28: Does it make sense to specify `Get` RPC as an option in the workflow described in this sentence?
```suggestion
Following a health check occurring, a caller uses the `List` or `Get` RPC to discover the
```

healthz/README.md
```diff
@@ -7,41 +14,239 @@ in distributed systems.
...
+artifacts that are assoicated with a component and its corresponding
```
hellt, line 29:
```suggestion
artifacts that are associated with a component and its corresponding
```

healthz/README.md
```diff
@@ -7,41 +14,239 @@ in distributed systems.
...
+subcomponents. Each artifact reflects an entity that is required to debug or
+further root cause the fault that occurs with it.
+
+The `Artifact` RPC is used to retrieve specific artifacts that are listed by
+the target system in the `List` RPC. Once retrieved each artifact can be
```
hellt, line 34:
```suggestion
the target system in the `List` or `Get` RPC. Once retrieved each artifact can be
```

healthz/README.md
hellt, line 29: It seems confusing to use the `artifact` term when we describe the `ComponentStatus`, as the artifact is a message encompassed in the status.
Later down the section we say that we acknowledge artifacts:

```
Once retrieved each artifact can be 'acknowledged' by the client of the RPC.
```

which confuses me, as we ack the `ComponentStatus` (which is also referred as `event` sometimes in this doc).

Shall we rather say `ComponentStatus` when describing the set of health status messages returned by `Get`/`List` RPCs?

healthz/README.md
```diff
@@ -7,41 +14,239 @@ in distributed systems.
...
-component/healthz/last-unhealthy = Timestamp
-component/healthz/unhealthy_count = Int64
+on components and if "unhealthy" gather implementor specific data to help
+triage or reproduce issues.
+
+## Design Background
+
+`Healthz` provides a means by which a user may initiate checks of health
+(through the `Check` RPC) or a system may report the results of a check that it
+has initiated of its own accord.
```
hellt, line 26: re "a system may report the results of a check that it has initiated of its own accord.":

1. when a system boots, should the relevant component manager perform initial health checks and thus generate ComponentStatus for components? 
2. When a certain component changes its health status from one state to another (think a tranceiver goes from healthy to unhealthy), should the relevant component manager record a ComponentStatus event with a distinct ID automatically?

marcushines: i would say it should look like this:

system boots
gnmi tree is initialized for things which have not run a check yet - they start in the unspecfied state.
During initialization of the subcomponents the system should perform initial health checks to put the components into a known state 
if the component is unhealthy - generate an event
if the components is healthy - set state and move on

2. yes if something goes unhealthly there should be an event - but how much work is done to "triage" that event is left to the vendor.

healthz/README.md
```diff
@@ -7,41 +14,239 @@ in distributed systems.
...
+subcomponents. Each artifact reflects an entity that is required to debug or
+further root cause the fault that occurs with it.
+
+The `Artifact` RPC is used to retrieve specific artifacts that are listed by
+the target system in the `List` RPC. Once retrieved each artifact can be
+'acknowledged' by the client of the RPC. Acknowledged artifacts are no longer
+returned in a list of the corresponding artifacts, and a device may use this
+status as a hint to allow garbage collection of artifacts that are no longer
+relevant. The device itself is responsible for garbage collection any may,
+if necessary, garbage collect artifacts that are not yet acknowledged.
```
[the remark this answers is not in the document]
marcushines: they should not be cleared just because the gnoi server restarted

healthz/README.md
```diff
@@ -7,41 +14,239 @@ in distributed systems.
...
+articacts to be reported by the device.
+
+
+## Architecture
+
+### Use of gNMI paths
+
+Where a gNMI path (`gnoi.types.Path`) is used in `Healthz`, the path specified
+should be the complete path to a specific component - i.e.,
+`/components/component[name=FOO]`.
```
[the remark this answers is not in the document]
robshakir: BGP itself would not necessarily be exposed as a component, but it is valid to have software processes in the `/components` tree. Based on this, I don't _think_ that there is a need to restrict this if the implementation supports a health check on software processes.

healthz/README.md
robshakir: I agree -- I've tried to standardise that "event" means a `ComponentStatus`. I think that this is clearer, since `ComponentStatus` itself seems obsure from a human understanding. PTAL (will push this change after sending this comment and addressing the others).

healthz/README.md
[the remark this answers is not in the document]
robshakir: Added a clarification -- I think generally these need to persist across component restarts too, since healthz information that corresponds to something like a rebooting linecard should not be removed as the LC reboots IMHO.

healthz/README.md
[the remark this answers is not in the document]
hellt: @robshakir okay, my reading here is that it is not prescribed by healthz (or Openconfig) to expose software services under `/components`, moreover, the schema for `/components` container hints that it is meant to be used for HW components as it has sub-containers like chassis, CPU, etc.

Therefore, it seems okay to first focus on the hardware parts in the early days of healthz and approach SW components later

healthz/README.md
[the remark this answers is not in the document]
hellt: so essentially we have to persist the events until they are garbage collected/acked.

healthz/README.md
```diff
@@ -7,41 +14,243 @@ in distributed systems.
...
+
+## Design Background
+
+`Healthz` provides a means by which a user may initiate checks of health
+(through the `Check` RPC) or a system may report the results of a check that it
+has initiated of its own accord. 
+
+Following a health check occurring, a caller uses the `List` or `Get` RPC to discover the
+health "events" (expressed as a `ComponentStatus` message) that are assoicated
+with a component and its corresponding subcomponents. Each event reflects an
```
hellt, line 30:
```suggestion
with a component and its corresponding subcomponents. Each event reflects a
```

healthz/README.md
```diff
@@ -7,41 +14,243 @@ in distributed systems.
...
+
+#### Exporting internal or specific debug logs
+
+For a component in the system the developers could provide a debug log for that
+specific component or other trace / dump information.  The `Check()` and
+`Artifact()` APIs could be used in these cases to create the setup and then
+download the specific data in a secure, reliable way (rather than a general file
+scp). If the device can "automatically" generate these specific
+data the `Get` API can also be used to check for existence and to download the
+artifacts. Note that to support this use case, asdditional gNMI paths are likely required to
```
hellt, line 237:
```suggestion
artifacts. Note that to support this use case, additional gNMI paths are likely required to
```

healthz/README.md
```diff
@@ -1,4 +1,11 @@
-# Healthz Overview
+# gNOI `Healthz` Streaming RPC Design
+**Contributors**: hines@google.com, robjs@google.com, bneville@arista.com, roman.dodin@nokia.com
+**Last Updated**: 2022-11-02
+
+## Background
+
+ * [gNOI Repository](https://github.com/openconfig/gnoi)
+ * [gNOI `Healthz` service](https://github.com/openconfig/gnoi/tree/master/healthz)
```
hellt, line 8: linting: remove extra space in the unordered list
```suggestion
* [gNOI Repository](https://github.com/openconfig/gnoi)
* [gNOI `Healthz` service](https://github.com/openconfig/gnoi/tree/master/healthz)
```

healthz/README.md
```diff
@@ -7,41 +14,243 @@ in distributed systems.
...
+#### Collect a snapshot of a particular component.
+
+For components which can provide a "snapshot" of state the healthz artifact
+endpoint can be provided to take a snapshot of state (ideally serialized as a
+proto or other typed data structure) and provided to the caller. 
+
+Examples:
+
+ * Get optical data from i2c bus on optical systems
+ * Get system database state for the core pub/sub of a device
```
hellt, line 227:
```suggestion
* Get optical data from i2c bus on optical systems
* Get system database state for the core pub/sub of a device
```

healthz/README.md
```diff
@@ -7,41 +14,243 @@ in distributed systems.
...
+
+A call to the `Check` endpoint will allow for "very expensive" debugging
+commands such as causing the device to pause and snapshot its system database
+or to examine the state of a protocol.  These commands will likely be
+considered "service impacting" and should have a clear security ACL restricting
+their use during normal operations of the device.
+
+## User Experience
+
+#### A BGP routing process goes unhealthy due to a crash.
```
hellt, line 167: next level heading should be level 3 `###`
```suggestion
### A BGP routing process goes unhealthy due to a crash.
```

healthz/README.md
```diff
@@ -7,41 +14,243 @@ in distributed systems.
...
+
+##### Expected Action
```
hellt, line 169:
```suggestion
#### Expected Action
```

healthz/README.md
```diff
@@ -7,41 +14,243 @@ in distributed systems.
...
+#### A chassis linecard goes unhealthy due to hardware failure.
+
+##### Expected Action
+
+Logs / dumps collected and the gNMI components are marked unhealthy.
+
+A call to Healthz.Get should respond with the current state of the component
+and provide feedback that there is an artifact available to be requested.
+
+A Healthz call should stream the core dump over the service back to the caller.
```
hellt, line 190:
```suggestion
A `Healthz.Artifact()` call should stream the core dump over the service back to the caller.
```

healthz/README.md
```diff
@@ -7,41 +14,243 @@ in distributed systems.
...
+
+A Healthz call should stream the logs over the service back to the caller.
```
hellt, line 192:
```suggestion
A `Healthz.Artifact()` call should stream the logs over the service back to the caller.
```

healthz/README.md
```diff
@@ -7,41 +14,243 @@ in distributed systems.
...
+#### A BGP routing process goes unhealthy due to a crash.
+
+##### Expected Action
+
+Write the core dump to a file system.
+
+A call to `Healthz.Get` should respond with the current state of the component
+and provide feedback that there is an artifact available to be requested.
+
+A Healthz call should stream the core dump over the service back to the caller.
```
hellt, line 176:
```suggestion
A `Healthz.Artifact()` call should stream the core dump over the service back to the caller.
```

healthz/README.md
```diff
@@ -7,41 +14,243 @@ in distributed systems.
...
+
+The component that is specified should correspond to the software process that
+implements the BGP protocol in the system.
```
hellt, line 179: I think it would be very cool to augment this process with a few more details, which would clarify the interopretations between healthz service and telemetry paths.

1. shall we add here a step after the core dump is written that goes like: 
    > A Healthz event MUST be generated stating the new health state is UNHEALTHY. It is left for the implementor to decide which artifacts are generated during this automatic event generation following the rule that those automatic event generations SHOULD NOT affect the performance of a system. If the triage of this failure requires substantial computational resource, `Healthz.Check()` must be used by the client.
2. Clarify, that `Healthz.Get()` might not list (all?) the artifacts, if collection of those artifacts was considered to be resource intensive task. In such case, the `Check()` RPC should be used to populate the list of available artifacts.
3. Add a note, that telemetry paths are updated automatically, without calling any healthz RPCs denoting the changed health state

hellt (commented):
Hi @robshakir @marcushines 
I have suggested a few clarifications that you might consider valid.

PS. plus some linter/typo work

healthz/README.md
hellt: Hi @robshakir 
can you ptal at this comment? The rest LGTM (with a small term glitch [here](https://github.com/openconfig/gnoi/pull/106#discussion_r1012733954))

healthz/README.md
```diff
@@ -7,41 +14,243 @@ in distributed systems.
...
+```protobuf
+rpc List(ListRequest) returns (ListResponse) {}
+```
+
+`List` returns all events that are associated with a particular component. In
+contrast to `Get`, it returns all events that the target has for the device. By
+default, events that are acknowledged are not returned.
+
+Similarly to `Get`, a series of `ComponentStatus` messages are returned with
+the same semantics as are described above.
```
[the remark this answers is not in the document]
robshakir: Made a slightly wider change.

healthz/README.md
```diff
@@ -7,41 +14,243 @@ in distributed systems.
...
+health "events" (expressed as a `ComponentStatus` message) that are assoicated
+with a component and its corresponding subcomponents. Each event reflects an
+collection of data that is required to debug or further root cause the fault
+that occurs with an entity in the system.
+
+The `Artifact` RPC is used to retrieve specific artifacts that are listed by
+the target system in the `List` or `Get` RPC. Once retrieved each artifact can be
+'acknowledged' by the client of the RPC. Acknowledged artifacts are no longer
+returned in a list of the corresponding artifacts, and a device may use this
+status as a hint to allow garbage collection of artifacts that are no longer
```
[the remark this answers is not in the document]
robshakir: Done.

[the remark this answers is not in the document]
hellt: I don't see changes to this paragraph in [79095d3](https://github.com/openconfig/gnoi/pull/106/commits/79095d3f6a11d9db4e60e4c282144c298a3bd563). But I am fine with merging this as is to not let this one stall

healthz/README.md
robshakir: I think this makes sense -- from a "safe" implementation perspective. Just looking at the place to add this to the doc.

healthz/README.md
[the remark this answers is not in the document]
robshakir: Sorry - just waiting to push it whilst working on the other item. Will have it this afternoon.

healthz/README.md
robshakir: Added a paragraph that basically says exactly what you say further up in the document since I think it is important for the designers of clients too.

healthz/README.md
```diff
@@ -7,41 +14,270 @@ in distributed systems.
...
+has initiated of its own accord.
+
+Following a health check occurring, a caller uses the `List` or `Get` RPC to
+discover the health "events" (expressed as a `ComponentStatus` message) that
+are associated with a component and its corresponding subcomponents. Each event
+reflects a collection of data that is required to debug or further root cause
+the fault that occurs with an entity in the system.
+
+The `Artifact` RPC is used to retrieve specific artifacts that are listed by
+the target system in the `List` or `Get` RPC.  Once retrieved an event, which
```
hellt, line 35: NIT: is this extra space needed? 
```suggestion
the target system in the `List` or `Get` RPC. Once retrieved an event, which
```

healthz/README.md
```diff
@@ -7,41 +14,270 @@ in distributed systems.
...
-  /component in the OC tree. The Get method is responsible for getting the
-  current status of the requested component path. This will cause the system
-  to fetch the current state of component this may entail running specific
-  validations and gathering any relevant status of the component to reflect
-  the "health check"
+1. A component of the system becomes unhealthy. The device MUST generate
+   a health state via telemetry that indicates that the `healthz/state/status`
+   of the component has transitioned to `UNHEALTHY`. It is left for the device
+   implementor to determine which artifacts are generated automatically by the
+   system immediately. The artifact collection SHOULD NOt impact the performance
```
hellt, line 89:
```suggestion
   system immediately. The artifact collection SHOULD NOT impact the performance
```

hellt (approved):
Thanks @robshakir 
Looks good to me, I spotted a few nit typos, but content-wise it looks great

healthz/README.md
```diff
@@ -7,41 +14,270 @@ in distributed systems.
...
-  to fetch the current state of component this may entail running specific
-  validations and gathering any relevant status of the component to reflect
-  the "health check"
+1. A component of the system becomes unhealthy. The device MUST generate
+   a health state via telemetry that indicates that the `healthz/state/status`
+   of the component has transitioned to `UNHEALTHY`. It is left for the device
+   implementor to determine which artifacts are generated automatically by the
+   system immediately. The artifact collection SHOULD NOT impact the performance
+   of the system.
+2. The responding client system observes the transition to `UNHEALTHY` and can
```
marcushines, line 91: this is an interesting idea
We would need to add an attribute which could signal there is "additional" work to be done to generate the artifacts 

It is easy to say "show tech" at the chassis is expensive but really we should never be getting "chassis" level checks we should really be only getting the needed artifacts for specific events

You do raise a reasonable use case though - let's say a system database reports an error - instead of pausing the system which could be service impacting to dumb the db- you could mark the event get the non service impacting data then leave a mark on the event that says - if you want more you can call check(event id) and it will go generate the other data and link to id which could be done after the device is drained or otherwise safe to get more data... That seems reasonable not sure it is "required" as we don't have a use case exactly of that

robshakir: Agreed. I've added the `event_id` to the `CheckRequest` to indicate that artifacts should be collected for a previous event, and made the clarifications that we discussed.

@hellt  PTAL at this too.

marcushines (approved):

healthz/README.md
hellt: @robshakir where did you add the text/field about it? I can't seem to find new commits since [92516ec](https://github.com/openconfig/gnoi/pull/106/commits/92516ecaa5bea9611cda4c21d6a7f3166b66dd31)

robshakir: I'm sorry -- this was my fault. I'd not spotted that the push needed a merge. PTAL at https://github.com/openconfig/gnoi/pull/106/commits/8cd01e1000a03f86f4dbfec6e9fc40a0c0700d6c

healthz/README.md
```diff
@@ -185,6 +193,14 @@ or to examine the state of a protocol.  These commands will likely be
 considered "service impacting" and should have a clear security ACL restricting
 their use during normal operations of the device.
 
+The `CheckRequest` message includes an `event_id` field. When populated this
+indicates that the `Check` should be performed for an event that has already
+occurred within the system. The device should trigger artifact collection of
+those artifacts that were not automatically collected. A `CheckRequest` for
+a previous `event_id` MUST NOT overwrite previous artifacts that were collected
+at the time of the event. The artifacts that are collected MUST be returned
+in the artifact list for the event when reported by the `Get` or `List` RPCs.
```
hellt, line 202: @robshakir I wonder how this is supposed to work, given that (I think) the health of a component can't be checked for the past events as the state of the component may already be HEALTHY or the artifacts collected won't contain the same state that were relevant at the time of a  healthy->unhealthy  transition.

I am inclined to not stall the PR though, as currently in our system we don't anticipate having expensive artifact collection processes which would require `Check` to work on a past `event_id`...

hellt: synced up offline
copying for historical purposes

> you call Check and it gathers some artifact that is based on the current state. This might include historical logs in the case of dumping some database, or retrieving some coredump from the secondary control plane.
But there's no guarantee that it captured the state at the time. For that, we need the artifacts that are automatically created to be complete.

hellt (approved):

PR #108, opened 2022-11-16
Update generated protobufs.

robshakir:
```
commit 26968c9a1b55134a879310986c2fa92c389127c3
Author: Rob Shakir <robjs@google.com>
Date:   Wed Nov 16 11:23:54 2022 -0800

    Update protobufs, bump healthz version.
    
     * (M) **/*.pb.go
       - Regenerate protobufs.
     * (M) heaslthz/healthz.proto
       - Update version to reflect new leaf addition in healthz in #106.
```

Update generated code and bump `gnoi.Healthz` version to 1.3.0 after leaf addition that was in #106.


robshakir:
There is existing client/server code that references the stubs that are generated in this repo, and it is the convention that we check these in alongside the protobuf in the place it is defined. There are projects that do depend on not generating the `.pb.go` at compile time -- most open source Go projects do this (looks like there are >2M `.pb.go` files on [GitHub](https://github.com/search?l=Go&q=filename%3A%2A.pb.go&type=Code))

We can extend this and add other languages based on the demand for it (e.g., gNMI has the python bindings checked in), please open a PR if you would like other languages. Equally, we have checks in other repos that does ensure that the protos have been updated, which can be enabled here.

I don't quite understand the objection here. What is the downside other than we don't have the CI set up to ensure that there isn't drift?

marcushines:
i believe the benefits of having an easy go mod install solution for these protos outweight any negatives
Also if other languages request the same generation logic we could add those libraries

marcushines (approved):

hellt:
I am coming from the camp where we salute generated stubs provided upstream, as this allows us to rely on the commonly generated stubs and import them in the client/server implementation, knowing that others will use the same product of generation.

PR #111, opened 2023-01-06
installing the package in one command on all Supervisors.

ramesh700:
Install RPC installs on both Supervisors and returns cluster_validated in Validated message as true on success. It returns NOT_SUPPORTED_ON_BACKUP in InstallError when asked to install on standby supervisor only. Image is present after Install on both Supervisors or neither Supervisor. Such systems do not need to sync the package from the peer Supervisor.

Activate RPC activates on both Supervisors and returns cluster_activated as true in ActivateOK message on success. It returns NOT_SUPPORTED_ON_BACKUP in ActivateError message when asked to activate on standby supervisors only. Such systems run the package by rebooting the entire system and do not need to switchover to the peer Supervisor before rebooting a Supervisor.

Added change to request snapshot of image and config to support auto recovery for systems which support snapshot to secondary disk on each Supervisor. When post upgrade reboot early init failures cause the system to boot from the second disk, the system auto recovers to the previously running image and config. This is specified by InstallRequest snapshot_to_secondary_disk.

### DIFF
--- a/os/os.proto
+++ b/os/os.proto
@@ -65,14 +65,23 @@ service OS {
   //                              <-- [Validated|InstallError]
   //
   // On a dual Supervisor Target, only the Active Supervisor runs this gNOI
-  // Service. The Install RPC applies to the Active Supervisor unless
+  // Service. 
+  // On systems which support installing the entire system with
+  // one upgrade command, one install RPC is sufficient to install the package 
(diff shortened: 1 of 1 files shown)


ramesh700:
opened 112 instead after fixing CLA error.

PR #112, opened 2023-01-06
installing the package in one command on all Supervisors.

ramesh700:
Install RPC installs on both Supervisors and returns cluster_validated in Validated message as true on success. It returns NOT_SUPPORTED_ON_BACKUP in InstallError when asked to install on standby supervisor only. Image is present after Install on both Supervisors or neither Supervisor. Such systems do not need to sync the package from the peer Supervisor.

Activate RPC activates on both Supervisors and returns cluster_activated as true in ActivateOK message on success. It returns NOT_SUPPORTED_ON_BACKUP in ActivateError message when asked to activate on standby supervisors only. Such systems run the package by rebooting the entire system and do not need to switchover to the peer Supervisor before rebooting a Supervisor.

Added change to request snapshot of image and config to support auto recovery for systems which support snapshot to secondary disk on each Supervisor. When post upgrade reboot early init failures cause the system to boot from the second disk, the system auto recovers to the previously running image and config. This is specified by InstallRequest snapshot_to_secondary_disk.


os/os.proto
```diff
@@ -120,12 +131,18 @@ service OS {
   rpc Verify(VerifyRequest) returns (VerifyResponse);
 }
 
+// Systems which save image and config to secondary disk on the same Supervisor for
+// auto recovery from failed upgrade will snapshot to secondary disk if snapshot_to_secondary_disk
+// is true
```
samribeiro, line 136: Missing period.

os/os.proto
```diff
@@ -65,14 +65,23 @@ service OS {
   //                              <-- [Validated|InstallError]
```
samribeiro, line 65: with line 19 please update to:

// Dual Supervisor Target is supported, where the above process is executed once
// for each Supervisor or once for the whole Cluster.

os/os.proto
```diff
@@ -65,14 +65,23 @@ service OS {
...
   //
```
samribeiro, line 66: bump version in line 10 to 0.1.3

os/os.proto
```diff
@@ -253,14 +286,26 @@ message InstallError {
...
+    // when requested to install on standby Supervisor only.
+    NOT_SUPPORTED_ON_BACKUP = 9;
   }
   Type type = 1;
   string detail = 2;
 }
 
 // The ActivateRequest is sent by the Client to the Target to initiate a change
 // in the next bootable OS version that is to be used on the Target.
+// On systems which install the image on the entire system in one command,
```
samribeiro, line 303: On systems which install the image on the entire cluster in one command

samribeiro:
Thank you @ramesh700. How about:

// Dual Supervisor Target is supported, where the above process is executed once
// for the whole Target or once for each Supervisor. Targets that require the above
// process executed once per Supervisor will reply with the flag
// Verify->VerifyResponse->individual_supervisor_install set to true.

The flag needs to be added, but with this way there is no need to have the cluster activated flag and the extra complexity in the Install RPC. The installer can just assumes that the entire Target is upgraded with a single install command regardless of the number of Supervisors it has.

os/os.proto
ramesh700: thanks for review, I have updated to this.

os/os.proto
ramesh700: Done

os/os.proto
```diff
@@ -65,14 +67,22 @@ service OS {
...
-  // The Target MUST always attempt to copy the OS package between Supervisors
-  // first before accepting the transfer from the Client. The syncing progress
-  // is reported to the client with InstallResponse->SyncProgress messages.
+  // On systems which support installing the entire system with one upgrade
+  // command, package is available on all Supervisors or none. If the package
+  // is absent on the Active Supervisor, Target does not try to copy the package
+  // from the standby Supervisor.
+  // On other systems, the Target MUST always attempt to copy the OS package 
+  // between Supervisors first before accepting the transfer from the Client. The 
+  // syncing progress is reported to the client with InstallResponse->SyncProgress messages.
```
samribeiro, line 85: On a dual Supervisor Target which requires installing the entire system with one Install RPC, one Install RPC is sufficient to install the package on all Supervisors. 

On a dual Supervisor Target which requires one Install RPC per supervisor, the Install RPC applies to the Active Supervisor unless InstallRequest->TransferRequest->standby_supervisor is set, in which case it applies to the Standby Supervisor. The Target MUST accept any order of package installation.

On a dual Supervisor Target which requires installing the entire system with one Install RPC, the package MUST be made available on all Supervisors or none. If upon installation the package is absent from either Supervisor, the Target will transfer to both Supervisors overwriting the existing one.

On a dual Supervisor Target which requires one Install RPC per supervisor, the Target MUST always attempt to copy the OS package between Supervisors first before accepting the transfer from the Client. The syncing progress is reported to the client with InstallResponse->SyncProgress messages.

os/os.proto
```diff
@@ -94,6 +104,8 @@ service OS {
   //                              ...
```
samribeiro, line 104: line 98:

Scenario 4 - On a dual Supervisor Target which requires one Install RPC per supervisor, and when one of the Supervisors already has the OS package but the other Supervisor is the target of the Install:

os/os.proto
```diff
@@ -94,6 +104,8 @@ service OS {
...
   //                              <-- [Validated|InstallError]
   //
+  //		- This scenario is not applicable to systems which install the package on all Supervisors
```
samribeiro, line 107: can be removed as the comment above covers it

os/os.proto
```diff
@@ -120,12 +132,18 @@ service OS {
...
+// is true
 message InstallRequest {
   oneof request {
     TransferRequest transfer_request = 1;
     bytes transfer_content = 2;
     TransferEnd transfer_end = 3;
   }
+  // For auto-recovery, snapshot will save the image and config to secondary disk 
+  // on the same Supervisor.
+  bool snapshot_to_secondary_disk = 4;
```
samribeiro, line 146: Current image and config? This should happen regardless, so this flag is not needed. See lines 44 and 366.

os/os.proto
```diff
@@ -135,19 +153,26 @@ message TransferRequest {
...
   // system, it will respond with InstallResponse->SyncProgress. In this,
   // latter, case - the client does not need to transfer the OS image. This
   // value can also be set empty, in which case the OS package is forced
   // transferred to the Target. The Target MUST never validate that this value
   // matches the one in the InstallResponse->Validated message, that is the
   // Client's responsibility.
+  // Systems which install the image on all Supervisors in one command
+  // will either have the image on both Supervisors or neither Supervisor.
+  // Such systems do not need to sync the image from peer Supervisor and
+  // will respond with TransferReady.
```
samribeiro, line 168: A dual Supervisor Target which requires installing the entire system with one Install RPC, will either have the image on both Supervisors or neither Supervisor. Such Targets do not need to sync the image from peer Supervisor and MUST immediately respond with TransferReady

os/os.proto
```diff
@@ -135,19 +153,26 @@ message TransferRequest {
...
+  // will either have the image on both Supervisors or neither Supervisor.
+  // Such systems do not need to sync the image from peer Supervisor and
+  // will respond with TransferReady. 
   string version = 1;
 
   // For a Target with dual Supervisors setting this flag instructs the Target
   // to perform the action on the Standby Supervisor.
+  // Systems which install the package on all Supervisors will return an
+  // error code of NOT_SUPPORTED_ON_STANDBY if requested to install on standby
+  // only.
```
samribeiro, line 175: A dual Supervisor Target which requires installing the entire system with one Install RPC, MUST return an error code of NOT_SUPPORTED_ON_STANDBY if requested to install on standby.

os/os.proto
```diff
@@ -177,6 +202,10 @@ message TransferEnd {
 // The InstallResponse is used by the Target to inform the Client about the
 // state of the Install RPC. At any stage of the process the Target can reply
 // with an Error message which MUST terminate the stream.
+// Systems which install the image on all Supervisors in one command
+// will either have the image on both Supervisors or neither Supervisor.
+// Such systems do not need to sync the image from peer Supervisor and
+// will respond with TransferReady if the package is absent on the Supervisor.
```
samribeiro, line 208: A dual Supervisor Target which requires installing the entire system with one Install RPC, will either have the image on both Supervisors or neither Supervisor. Such systems do not need to sync the image from the peer Supervisor and MUST respond with TransferReady if the package is absent from any Supervisor.

os/os.proto
```diff
@@ -204,6 +233,8 @@ message TransferProgress {
 
 // The SyncProgress message signals the Client about the progress of
 // transferring the OS package between Supervisors.
+// Systems which install the package on all Supervisors do not need to
+// transfer the package between Supervisors and will not send SyncProgress.
```
samribeiro, line 237: A dual Supervisor Target which requires installing the entire system with one Install RPC, does not need to transfer the package between Supervisors and MUST not send SyncProgress.

os/os.proto
```diff
@@ -253,14 +284,24 @@ message InstallError {
     UNEXPECTED_SWITCHOVER = 7;
     // Failed to sync the transferred OS package to the standby Supervisor. The
     // detail value MUST have more information.
+    // On systems which install the package on all Supervisors in one command,
```
samribeiro, line 287: A dual Supervisor Target which requires installing the entire system with one Install RPC, MUST not return SYNC_FAIL.

os/os.proto
```diff
@@ -303,6 +349,9 @@ message ActivateError {
     // There is no OS package with the version requested for activation. This is
     // also used for an empty version string.
     NON_EXISTENT_VERSION = 1;
+    // Systems which activate on all Supervisors will return NOT_SUPPORTED_ON_BACKUP
+    // when asked to activate on standby Supervisor only.
+    NOT_SUPPORTED_ON_BACKUP = 2;
```
samribeiro, line 354: Dual Supervisor Target which requires installing the entire system with one Install RPC MUST return NOT_SUPPORTED_ON_BACKUP error when requested to Activate on standby Supervisor.

os/os.proto
```diff
@@ -321,6 +370,9 @@ message VerifyResponse {
   string activation_fail_message = 2;
 
   VerifyStandby verify_standby = 3;
+  // Targets that require the Install/Activate/Verify process executed once
+  // per supervisor reply with individual_supervisor_install set to true
+  bool individual_supervisor_install = 4;
```
samribeiro, line 375: Dual Supervisor Targets that require the Install/Activate/Verify process executed once per supervisor reply with individual_supervisor_install set to true

samribeiro (commented):
Thanks! A few comments.

os/os.proto
ramesh700: done

os/os.proto
samribeiro: this is missing

os/os.proto
samribeiro: this is missing

os/os.proto
samribeiro: this is missing

os/os.proto
```diff
@@ -7,7 +7,7 @@ import "types/types.proto";
 
 option go_package = "github.com/openconfig/gnoi/os";
 
-option (types.gnoi_version) = "0.1.2";
+option (types.gnoi_version) = "0.1.3";
```
samribeiro, line 10: lets bump it to 4 given that there is another push that will get in before this one and is already claiming 3.

samribeiro (commented):
Just a few notes more, thanks for the patience!

os/os.proto
ramesh700: done.

os/os.proto
ramesh700: done.

os/os.proto
ramesh700: done.

samribeiro:
LGMT, thanks!

samribeiro:
[os/os.proto:200:1] The line length is 81, but it must be shorter than 80
[os/os.proto:202:1] The line length is 83, but it must be shorter than 80
[os/os.proto:285:1] The line length is 81, but it must be shorter than 80

os/os.proto
```diff
@@ -253,14 +282,24 @@ message InstallError {
...
   Type type = 1;
   string detail = 2;
 }
 
 // The ActivateRequest is sent by the Client to the Target to initiate a change
 // in the next bootable OS version that is to be used on the Target.
+// Dual Supervisor Target which requires installing the entire system with 
+// one Install RPC MUST return NOT_SUPPORTED_ON_BACKUP error when requested 
+// to Activate on standby Supervisor.
+
```
hellt, line 302:
```suggestion
```

os/os.proto
```diff
@@ -204,6 +229,10 @@ message TransferProgress {
 
 // The SyncProgress message signals the Client about the progress of
 // transferring the OS package between Supervisors.
+// A dual Supervisor Target which requires installing the entire system with 
+// one Install RPC, does not need to transfer the package between Supervisors 
+// and MUST not send SyncProgress.
+
```
hellt, line 235:
```suggestion
```

os/os.proto
```diff
@@ -135,19 +147,27 @@ message TransferRequest {
...
   // latter, case - the client does not need to transfer the OS image. This
   // value can also be set empty, in which case the OS package is forced
   // transferred to the Target. The Target MUST never validate that this value
   // matches the one in the InstallResponse->Validated message, that is the
   // Client's responsibility.
+  // A dual Supervisor Target which requires installing the entire system 
+  // with one Install RPC, will either have the image on both Supervisors 
+  // or neither Supervisor. Such Targets do not need to sync the image 
+  // from peer Supervisor and MUST immediately respond with TransferReady
+
```
hellt, line 163:
```suggestion
```

os/os.proto
samribeiro: thanks!

samribeiro:
merge after #113

PR #114, opened 2023-01-25
[System] Clarify fields in the SwitchControlProcessorResponse

### LINKED ISSUE #93: Clarification around uptime in `SwitchControlProcessorResponse`
Hi all,

Could you please confirm what the provided value in the [uptime](https://github.com/openconfig/gnoi/blob/6bddacd72ef8aa06304b887df414a43c8e63ff27/system/system.proto#L99) field is meant to be? It seems to be useful to give the number of nanoseconds the to-be-switched-to supervisor has been operational - i.e. a delta between the last booted time and the current time, but the mention of epoch here seems to indicate it is meant to be either the current time or the last booted time.

The corresponding [featureprofile](https://github.com/openconfig/featureprofiles/blob/main/feature/gnoi/system/tests/supervisor_switchover_test/supervisor_switchover_test.go#L109) only tests that a value >0 is returned.

hellt:
This is to provide additional comments around the fields of the SwitchControlProcessorResponse.

The uptime comment has been changed to reflect the uptime nature of the field if that was the intention. For that reason, `since epoch` substring was removed, as epoch time indicates a point in time, not duration.

close #93 

PS1: I wonder if we need to clarify how to handle errors that might happen during the switchover? So far, rpc-sprcific errors are not part of the spec. 

PTAL @marcushines

### DIFF
--- a/system/system.proto
+++ b/system/system.proto
@@ -96,8 +96,10 @@ message SwitchControlProcessorRequest {
 
 message SwitchControlProcessorResponse {
   types.Path control_processor = 1;
-  string version = 2; // Current software version.
-  int64 uptime = 3;   // Uptime in nanoseconds since epoch.
+  // Current software version of the target Control Processor.
+  string version = 2;
+  // Uptime of the target Control Processor in nanoseconds.
+  int64 uptime = 3;
 }
 
 // A RebootRequest requests the specified target be rebooted using the specified
--- a/system/system.proto
+++ b/system/system.proto
@@ -91,10 +91,12 @@ service System {
 }
 
 message SwitchControlProcessorRequest {
+  // Path to the target Control Processor.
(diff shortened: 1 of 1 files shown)


system/system.proto
```diff
@@ -96,8 +96,10 @@ message SwitchControlProcessorRequest {
 
 message SwitchControlProcessorResponse {
   types.Path control_processor = 1;
```
xw-g, line 98: you might also call it out that the control_processor is about the target processor too?

hellt: @xw-g done in [7846102](https://github.com/openconfig/gnoi/pull/114/commits/784610276dad74a47c5d38d740882fb0efb73745)

marcushines (approved):

marcushines (approved):

PR #118, opened 2023-03-31
Target does zeroize while maintaining certificates.

ramesh700:


sachendras (approved):

marcushines (approved):

hellt:
@marcushines @sachendras we have to bump a service version when introducing changes. For us, implementors, this is crucial, as we have to declare compatibility with a certain version of a service.

PR #119, opened 2023-04-06
Add image name used in Install RPC to VerifyResponse and StandyRespon…

ramesh700:
…se. For images which are named

differently from the official image version name, returning image name allows for verifying the image passed in via the Install RPC is running.


sachendras:
This pull is not required anymore.

PR #121, opened 2023-04-17
Remove interface service based on no implementations nor no clear use

marcushines:
gNMI can set all of these configurations at this point This service no longer is needed
Also the reset counters actually breaks many streaming implementations


tomek-US (approved):
LGTM

dplore:
Added to agenda for April 18 openconfig operators meeting

dplore:
Reviewed without objection from the OC Operators meeting

dplore (approved):

charanjith-anet:
Hi,

"gNMI can set all of these configurations at this point"
Can you please give an example of a path+value that can (or is expected to) reset L2 interface counters?

Thanks

dplore:
Hi @charanjith-anet , this PR assumed that reset interface counters is not a needed feature and in fact can break some streaming implementations.  Resetting a counter is viewed as an interactive CLI feature and not a backend telemetry streaming interface.

PR #123, opened 2023-06-12
Initial commit of a gNOI containerz service.

alshabib:
This service enables transfer of containers to
a networks device as well as starting, stopping containers.


alshabib:
@robshakir Could you please have a look?

containerz/BUILD.bazel
```diff
@@ -0,0 +1,59 @@
+load("@rules_proto//proto:defs.bzl", "proto_library")
+load("@io_bazel_rules_go//go:def.bzl", "go_library")
+load("@io_bazel_rules_go//proto:def.bzl", "go_proto_library")
+
+#Copyright 2021 Google LLC
```
robshakir, line 5:
```suggestion
# Copyright 2023 Google LLC
```

containerz/containerz.proto
```diff
@@ -0,0 +1,286 @@
...
+syntax = "proto3";
+
+package gnoi.containerz;
+
+import "common/common.proto";
+import "google/rpc/status.proto";
+import "types/types.proto";
+
+option go_package = "github.com/openconfig/gnoi/containerz";
+
```
robshakir, line 28: Should we add a version tag here (like [this](https://github.com/openconfig/gnoi/blob/main/file/file.proto#L26))? I think it might also be worth adding a comment that this is a first pass at such an implementation.

robshakir (approved):
LGTM, thanks.

PR #128, opened 2023-08-04
Update CI to run go 1.19 + go 1.20

robshakir:
Update versions in `go.yml` to run latest versions of Go.


marcushines (approved):

wenovus (changes requested):
This already is doing the actions that have been added:
https://github.com/openconfig/common-ci/blob/8f3db83e5548415f4f3999e42dd5f9bf7a203837/.github/workflows/basic_go.yml#L10

`oldstable` and `stable` are the latest two versions of Go, which is exactly what Go supports. Therefore 1.21 and 1.20 are already supported.

If other versions are needed, then can use this flag to override the defaults: https://github.com/openconfig/kne/blob/d973219c357631e67745329985c2771499e6181c/.github/workflows/go.yml#L18

robshakir:
I think this predated the commonci workflow being added. I'll close it since I don't think we need it right now.

PR #135, opened 2023-10-13
Update types.proto

marcushines:
Add clarification on the hash bytes value content

### DIFF
--- a/types/types.proto
+++ b/types/types.proto
@@ -45,7 +45,7 @@ message HashType {
     MD5 = 3;
   }
   HashMethod method = 1;
-  // hash is the raw byte string of the hash function.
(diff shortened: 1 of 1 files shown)


thesrinath (approved):

thesrinath (approved):

types/types.proto
```diff
@@ -45,6 +45,8 @@ message HashType {
     MD5 = 3;
   }
   HashMethod method = 1;
+  // hash is the raw byte string of the hash function.
```
robshakir, line 48:
```suggestion
  // hash is the raw byte string returned by the hash function.
```

robshakir (approved):

PR #136, opened 2023-10-17
Clarify RebootMethod

dplore:


system/system.proto
```diff
@@ -118,6 +118,8 @@ message RebootResponse {
 // A RebootMethod determines what should be done with a target when a Reboot is
 // requested.  Only the COLD method is required to be supported by all
 // targets.  Methods the target does not support should result in failure.
+// 
+// If a target does not support a method, `INVALID_ARGUEMENT` should be returned.
```
greg-dennis: ARGUEMENT->ARGUMENT

greg-dennis: Is this line redundant with the last line of the preceding paragraph?

system/system.proto
```diff
@@ -118,6 +118,9 @@ message RebootResponse {
 // A RebootMethod determines what should be done with a target when a Reboot is
 // requested.  Only the COLD method is required to be supported by all
 // targets.  Methods the target does not support should result in failure.
+// 
+// If a target does not support a method, `INVALID_ARGUEMENT` should be
```
greg-dennis, line 122: Should i also say that INVALID_ARGUMENT should be returned if the default/unknown method is provided or is that implicit?

greg-dennis (changes requested):

system/system.proto
dplore: fixed

dplore: fixed

system/system.proto
dplore: Added explicity reference for both scenarios.

greg-dennis (approved):

dplore:
I'll wait a bit for any other comments and merge later today

ejbrever (approved):
LGTM

PR #139, opened 2023-10-31
Adds auto_restart to containerz.StopRequest.

tomek-US:
As there is no containerz.Restart() RPC it is possible to loose access to the device after calling Stop() on the container implementing the gNOI.containerz service.

Adding the auto_restart field to the Stop() RPC solves this problem.


containerz/containerz.proto
```diff
@@ -258,6 +258,10 @@ message StopRequest {
 
   // If force is set, the target should attempt to forcibly kill the container.
   bool force = 2;
+
+  // If auto_restart is set, the target should start the container immediately
+  // after stopping it.
+  bool auto_restart = 3;
```
marcushines, line 264: bool restart

i think auto_restart makes it feel like the container will always restart

tomek-US: Make sense. Thx for making the change!

marcushines (approved):

PR #140, opened 2023-11-01
Create dependabot.yml

marcushines:


.github/dependabot.yml
```diff
@@ -0,0 +1,15 @@
+ To get started with Dependabot version updates, you'll need to specify which
+# package ecosystems to update and where the package manifests are located.
```
wenovus, line 2: I think the hash tag was removed accidentally here.

wenovus (changes requested):

wenovus (approved):

PR #147, opened 2023-11-03
Create Debug gNOI service for interacting with devices for debugging

marcushines:
issues with there is no appropriate OC or other service model.


debug/README.md
```diff
@@ -0,0 +1,40 @@
+# gNOI `Debug` Streaming RPC Design
```
robshakir, line 1: (nit) please wrap to 80ch

debug/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+**Contributors**: hines@google.com, robjs@google.com
+**Last Updated**: 2023-11-04
+
+## Background
+
+* [gNOI Repository](https://github.com/openconfig/gnoi)
+* [gNOI `Debug` service](https://github.com/openconfig/gnoi/tree/master/debug)
+
+For all legacy devices that provided a CLI on box, providers  have leveraged this CLI through services to provide users with the ability to interact with device CLI's to both get information as well as set operational state on devices.  This interaction is very vendor specific and requires significant overhead to maintain the vendor specific bindings throughout the operational life cycle of a device.
```
robshakir, line 11: This doesn't parse that well. Suggest:

"Traditional network devices provide a CLI, which providers have leveraged for configuration and retrieval of operational state. The implementation of these CLIs is vendor-specific, and requires specific per-vendor client code and logic. Programmatic use of the CLI is fragile -- being based on parsing unstructured strings designed for human consumption, rather than structured machine-friendly data."

debug/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+With the introduction of g* services, the goal has been to remove vendor specific data formats from the view operators.  This lets operators have standard models for interacting with any number of vendor devices consistently. There however are gaps between those endpoints versioning and the ability to troubleshoot specific data on a device before API's can be updated.  This proposal is to enable a lightweight interface via gRPC to still access shell level interactions on the device in a secure, maintainable way.
```
robshakir, line 13:
```suggestion
With the introduction of g* services, the goal has been to remove vendor specific data formats from the view operators.  This lets operators have standard models for interacting with any number of vendor devices consistently. There however are gaps between those endpoints versioning and the ability to troubleshoot specific data on a device before APIs can be updated.  This proposal is to enable a lightweight interface via gRPC to still access shell level interactions on the device in a secure, maintainable way.
```

debug/README.md
```diff
@@ -0,0 +1,40 @@
...
+
```
robshakir, line 14: Suggest to add here:

"It is expected that where structured data is identified to be useful for an operator the suite of g* services, and/or underlying data models (e.g., OpenConfig) are extended to cover these use cases. However, it is critical to recognise that these APIs are not the "native" device APIs, and hence upgrading them requires some new software to be introduced."

debug/debug.proto
```diff
@@ -0,0 +1,77 @@
...
+  // command. (e.g. Cli modes must exit after sending a command)
+  // Errors:
+  //  InvalidArgument: for unspecified mode
+  rpc Debug (DebugRequest) returns (stream DebugResponse);
+}
+
+message DebugRequest {
+  enum Mode {
+    MODE_UNSPECIFIED = 0;
+    MODE_SHELL = 1;
```
robshakir, line 43:
```suggestion
    MODE_SHELL = 1; // Run the command using a standard Unix shell.
```

debug/debug.proto
```diff
@@ -0,0 +1,77 @@
...
+    MODE_CLI = 2;
```
robshakir, line 44:
```suggestion
    MODE_CLI = 2; // Run the command using the vendor specific native CLI.
```

debug/debug.proto
```diff
@@ -0,0 +1,77 @@
...
+  }
+  // Mode the commands will be executed in.
+  Mode mode = 1;
+  // Raw bytes for the command to be executed.
+  bytes command = 2;
```
robshakir, line 49: Why did we make this `bytes`? Is there a case where it's ever something other than a string? I'm wondering whether having `bytes` here introduces any new attack vector vs. having a string that could be parsed out (I presume I can't just put an statically linked executable in this field ;-))

marcushines: i could do string i think it won't matter

robshakir: After our discussion, I'm OK with `bytes`.

robshakir (approved):

debug/README.md
```diff
@@ -0,0 +1,62 @@
...
+logic. Programmatic use of the CLI is fragile -- being based on parsing
+unstructured strings designed for human consumption, rather than structured
+machine-friendly data.
+
+With the introduction of g* services, the goal has been to remove vendor
+specific data formats from the view operators. This lets operators have
+standard models for interacting with any number of vendor devices consistently.
+There however are gaps between those endpoints versioning and the ability to
+troubleshoot specific data on a device before APIs can be updated. This
+proposal is to enable a lightweight interface via gRPC to still access shell
```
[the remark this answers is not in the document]
marcushines: hehe yep :)

debug/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+It is expected that where structured data is identified to be useful for an
+operator the suite of g* services, and/or underlying data models
+(e.g., OpenConfig) are extended to cover these use cases. However, it is
+critical to recognise that these APIs are not the "native" device APIs, and
+hence upgrading them requires some new software to be introduced.
+
+## Architecture
+
+The service will run on a specified port. This service upon recieving a command
```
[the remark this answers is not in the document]
marcushines: it will depend on the vendor - given this doesn't look like gnoi/gnmi i am not sure it should be generally in the same service

PR #160, opened 2024-01-30
gnoi: add service for extension management

brianneville:
Add the Extension service to allow for network operators to manage extensions on the system, where an extension is some arbitrary patch or upgrade which can be applied to services running on the system.

These extensions can have signatures which may be verified (e.g. a signature may identify an extension as originating from the vendor).

The extension.proto file details proposed workflows for managing extensions (including tranferring those extensions onto the system, configuring signature verification options, and install/uninstall workflow)


extension/extension.proto
```diff
@@ -0,0 +1,174 @@
...
+
+option go_package = "github.com/openconfig/gnoi/extension";
+
+option (types.gnoi_version) = "0.1.0";
+
+// Extension is a service for managing extensions on the system.
+// Only one RPC (SetSignatureVerification, FinalizeExtensions,
+// InstallExtension, UninstallExtension)
+// may be in progress at any given time
+service Extension{
```
hellt, line 15: what is meant by an `extension`? Is it a software bundle that runs on a NOS?
The term seems to be not so well known...

brianneville: an extension here is just an bundle of arbitrary files (e.g. RPMs) which can be installed and applied over the system to modify the behaviour of some services on the system. 
This is very generic, theres nothing specific here about what an extension must do or any requirements for the extension file itself. 
I've added a comment in the next commit https://github.com/openconfig/gnoi/pull/160/commits/ea29e019b9b3d985d5b3472140f60aed1d8c4106

As an example of this: 
Some network operator experiences a bug with BGP services, and a vendor may provide that network operator with an extension which can be installed to apply a fix for that bug. The network operator transfers the extension (verifying that it was signed by the device manufacturer if desired) and then activates it, which applies the fix to BGP.
 
Network operators may prefer doing this rather than performing a full upgrade to a new OS version using `gnoi OS`, as an extension can be applied without installing a new software image, and it doesnt affect any unrelated services

extension/extension.proto
```diff
@@ -0,0 +1,177 @@
...
+message TransferRequest{
+  oneof request {
+    // name of the extension  (including the file extension)
+    string extension_name = 1;
+    // raw byte contents to be streamed into the extension file.
+    bytes contents = 2;
+    // hash of the file.
+    types.HashType hash = 3;
+  }
+}
```
hellt, line 177: I wonder if we want to have multiple file transfer ops spread across different services.

System.Package - https://gnxi.srlinux.dev/gnoi/system#SetPackageRequest looks to be a close match

extension/extension.proto
sulrich: @hellt - 

the notion of extension installation is pretty common across NOS.  this might take the form of vendor-provided (and signed) extensions or the installation of common linux utilities that extend the base capabilities of the NOS.

examples include: 
- juniper/HPE junos(-evolved) - [How to Install Third-Party Software on Devices Running Junos OS Evolved | Junos OS Evolved | Juniper Networks](https://www.juniper.net/documentation/us/en/software/junos/junos-install-upgrade-evo/topics/concept/junos-evolved-install-third-party-software-overview.html)
- cisco ios-xr - [Cisco IOS XR Setup and Upgrade Guide for Cisco 8000 Series Routers - Manage the Router \[Cisco 8000 Series Routers\] - Cisco](https://www.cisco.com/c/en/us/td/docs/iosxr/cisco8000/b-setup-and-upgrade-cisco8k/manage-software.html)
- arista [EOS 4.31.1F - Managing EOS Extensions - Arista](https://www.arista.com/en/um-eos/eos-managing-eos-extensions?searchword=eos%20section%206%207%20switch%20administration%20commands)

hellt: @sulrich yes, the concept is clear, it is just that "extension" seems to be a peculiar choice for a package.

But maybe it is just me :)

extension/extension.proto
```diff
@@ -0,0 +1,177 @@
...
+// which are installed to modify the behaviour of the services on the
+// system.
+// Only one RPC (SetSignatureVerification, FinalizeExtensions,
+// InstallExtension, UninstallExtension)
+// may be in progress at any given time
+service Extension{
+
+  // SetSignatureVerification sets the methods for which the
+  // system should verify incoming extensions
+  rpc SetSignatureVerification(SigVerificationRequest)
```
robshakir, line 22: What's the motivation for this being a separate RPC which then implies there's some state held across requests? It seems like it could just be a flag to `Install...`.

extension/extension.proto
```diff
@@ -0,0 +1,177 @@
...
+  // (note, other fields may be set )          -->
+  // .. server installs extension according to fields.
+  //                                           <-- InstallExtensionResponse
+  //                              (with acknowledgement = INSTALL_COMPLETE)
+  // STREAM END
+  // Note that all installations will install the extension, but that
+  // extension will not be applied until the affected services are reloaded.
+  // To reload services after installing extensions, issue a
+  // FinalizeExtensions RPC after finishing the InstallExtension RPC.
+  rpc InstallExtension(stream InstallExtensionRequest)
```
robshakir, line 79: I think someone else observed this in another comment, but there is `gnoi.System.SetPackage` which was designed to be a "generic" means to install a package on a system. `gnoi.os.Install` was created specifically because there were a lot of semantics around NOS installation that were beyond `SetPackage`.

Here, what's the motivation that separates this RPC from `SetPackage`? Do you envisage that we deprecate `SetPackage` if this is merged?

brianneville: I'll confess to not knowing exactly what `System.SetPackage` is supposed to be used for given the current gNOI services.
from the description:
https://github.com/openconfig/gnoi/blob/f62dc1b711ef0809752792917fb556824fbcd013/system/system.proto#L53-L54
It seems to be a very generic way to download packages, where i dont think its very clear what a package is (this may be intentional - perhaps a package is just any bundle of files).  

The comment alludes to the RPC being potentially used to download bootable images, however this functionality was later implemented in the gNOI OS package, with this comment from the PR explaining how `SetPackage` is too generic to really capture the nuances with OS installation: https://github.com/openconfig/gnoi/pull/19#issuecomment-479083625
I'd argue that the same is true here with extensions. 
There's more to installing an extension than just file transfer and "activation" (which is really all that the `SetPackage` RPC affords). 

Some examples of differences here between `Extension` RPCs and `SetPackage`:
- `SetPackage` only allows for activation of a package after transfer, it has no capability to "activate" a package which is already present on the device
- `SetPackage` only has one option, either to "activate" a package or not:
https://github.com/openconfig/gnoi/blob/f62dc1b711ef0809752792917fb556824fbcd013/system/system.proto#L320-L323
There are other options which are relevant to extension installation, related to signature verification, dependency checks, whether this extension should be continuously applied after reboots. 
This is specific to installing an `extension` and not just something that would be relevant to any generic package.
- `Extension` allows for the installation/activation workflow to be spread over multiple RPCs (`InstallExtension` and `FinalizeExtensions`, which gives an opportunity for a network operator to install multiple extensions in sequence, verifying that they all succeed before finalizing and reloading the relevant services. 


>Here, what's the motivation that separates this RPC from SetPackage? Do you envisage that we deprecate SetPackage if this is merged?

I would be ok with deprecating it, but im not sure how widely used it is or what other vendors/network operators use it for. 
Deprecating it would be a different PR/issue though, made as a followup after this PR

extension/extension.proto
brianneville: >I wonder if we want to have multiple file transfer ops spread across different services.

I am ok with having multiple file transfer ops across services, mainly because:
1. [gNSI Authz](https://github.com/openconfig/gnsi/tree/a80d61cd116210b0dfcac021ecd74906f3bbf4fa/authz) is service/RPC based. If we delegate that all file transfer operations must be done by some common RPC (e.g. `gNOI File/Put` RPC), then this means that for a system running gNSI Authz, all clients must also be authorized to run the `gNOI File/Put` RPC, which would enable them to put any file on the system that they want.
I think its better to build the file-transfer operations into the RPC itself, so that the a client created for the purpose of "transferring extensions onto the system" is only given the ability to do that, rather than general access to the whole filesystem. 

2. (smaller point) for ease of use, i like that a single client (one for only the `Extension` service) can be used to interact with a service, rather than creating multiple clients (one for `File` and another for `Extension`)

>System.Package - https://gnxi.srlinux.dev/gnoi/system#SetPackageRequest looks to be a close match

I mention this in a separate comment here: https://github.com/openconfig/gnoi/pull/160/#discussion_r1472784052

extension/extension.proto
brianneville: This was made a separate RPC as there is state being held across requests :)
the intention is that the signature verification method is global config (e.g. it may mark a single SSL profile as being the chosen profile to verify against). 
It can be set once before any extensions are installed, so that the same config is applied to all extensions installed afterwards (in the case of multiple `InstallExtension` RPCs), rather than relying on the client to re-specify the same option in their requests.

extension/extension.proto
robshakir: The spec here doesn't seem to really capture the nuances that you mention above that mean that `SetPackage` isn't the right solution. Can I suggest that we start with a brief design document that describes what we're trying to solve here -- particularly covering the primitives that you propose so that other implementors and operators can comment on this?

Note all the things that you mention about `SetPackage` could be done by augmenting its functionality to include the specific gaps. I don't know that is the right solution, but I also don't have a good picture about what workflows you're trying to achieve here -- which the design doc would help close.

brianneville: I've added a new commit with a README.md file outlining what an extension is and how the typical extension installation workflow would proceed, please take a look.

brianneville: Hi rob, 

Can you take a look at the README.md in this PR?  
It would be good to get this PR finalized

alshabib: Thanks @robshakir and thanks to @brianneville for adding the readme, managing extensions is something lacking in gnoi.

Would it make sense to split the transfer of an extension and its installation? For example, could we use `os.SetPackage` or `file.Put` to transfer the extension to the device and then this service could be dedicated to extension management. 

Other comments:

* Could the signature verification be made part of the Installation? Given that an individual Installation can opt out of the verification it would seem to make sense to specify the verification semantics as part of the Install.
* Would it be possible to add API for listing installed and/or activated extensions?

brianneville: >Would it make sense to split the transfer of an extension and its installation? For example, could we use os.SetPackage or file.Put to transfer the extension to the device and then this service could be dedicated to extension management.

I think having a separate way to transfer extensions onto the system is reasonable here, rather than reusing `file.Put`  for the reasons in the above comment https://github.com/openconfig/gnoi/pull/160#discussion_r1472785123 (i.e.  gNSI Authz is RPC based and we should be relucant to grant someone access to "uploading any files onto the system", and it simplifies installation if only one service client is required).

Additionally, if the client uses `file.Put`, then they'd need to know where to upload the extension on the system, rather than having that be abstracted away. (i.e. they need to know the vendor-specific location of where extension files should be placed on the device).

>For example, could we use os.SetPackage or file.Put to transfer the extension to the device and then this service could be dedicated to extension management.

I'm still not in favour of using `os.SetPackage`, for the reasons given in my comments above.
I really don't think its adequate in its current form, and i don't like the idea of overhauling that single RPC to try and force all this extension management functionality (as well as any future functionality) into it.
I dont think its a good idea to have to balance "all future extension-related changes" with "maintaining backwards compatability" for `os.SetPackage`. 
To me it seems more reasonable to just leave `os.SetPackage` alone, and develop this extension service in isolation.

 > Other comments:
>Could the signature verification be made part of the Installation? Given that an individual Installation can opt out of the verification it would seem to make sense to specify the verification semantics as part of the Install.

 yep this sounds fine 

>  Would it be possible to add API for listing installed and/or activated extensions?

Would it make more sense to add this information into some YANG model (since its just pure state for the system), so gNMI clients could subscribe to it or get it? 
If you think not though i can definitely add this here, having a way to list extension information is a good idea.

alshabib: Thanks Brian,

I think it would be good to support List extensions here initially and perhaps later adding this to gNMI if it makes sense.

Do you think there should be a third oneof type in InstallExtensionRequest to Activate extensions? I think there is a distinction between installation and activation or is the intent to activate the extension during the call to FinalizeExtensions?

alshabib: Ah understood - thanks for the clarification.

@earies @aaronmillisor @harveyyoung Would you be able to take a look at this and comment whether this is compatible with your respective NOSs?

itbwtw: Just trying to follow up here. Any comment on this? It'd be good if we can finalize this PR

ceneilson: I am also following up here, as we'd like to make progress internally. What will it take to get this over the finish line?

@alshabib if we are unable to get further input, can we look at approving this?

ceneilson: @robshakir , same question but directed to you I guess?

charanjith-anet: @alshabib @robshakir any feedback?

charanjith-anet: @earies @aaronmillisor @harveyyoung @alshabib @robshakir
Please review and approve everything looks good for you. We'd like to make progress.

robshakir (commented):
Looks like we are lacking a bit of other implementor feedback here, so I did some research:

 * [Cisco XR](https://www.cisco.com/c/en/us/td/docs/iosxr/ncs5xx/system-setup/24xx/b-system-setup-cg-24xx-ncs540/m-install-owner-and-partner-rpms-using-xr-infrastructure.html) supports installing RPMs, and has some similar flags. Currently, it looks like there is not support for `load_after_boot` or configurable signature IDs.
 * [SRLinux](https://learn.srlinux.dev/ndk/guide/agent-install-and-ops/) provides means to install `.deb` files -- which gives a means to be able to use `dpkg` to install these -- which implies some support of existing options here. `load_after_boot` seems likely to be missing here too.
 * [JUNOS evo](https://www.juniper.net/documentation/us/en/software/junos/junos-install-upgrade-evo/topics/concept/junos-evolved-install-third-party-software-overview.html#:~:text=You%20install%20the%20third%2Dparty,of%20the%20active%20Routing%20Engine.&text=Note:,a%20third%2Dparty%20application%20fails) has `request system software` that covers installing software with similar options.

I think the argument that there's some post-install/signature verification required that `SetPackage` does not give us seems reasonable to me. I think we should look at whether we can just deprecate `SetPackage` going forward.

I'll initiate a discussion on this next week and then we can see whether we can last-call this.

extension/README.md
```diff
@@ -0,0 +1,142 @@
...
+- Activation may affect the gNOI service, if the program running
+  the gNOI service needs to reload.      
+  SetPackage doesn't have any expectation that the RPC could be interrupted
+  due to activation.
+- SetPackage allows for packages to be transferred to arbitrary locations
+  on the filesystem, while this should be abstracted away from the client
+  when installing extensions to allow for the system to group all extension
+  files under a single directory (the system shouldn't search its entire
+  filesystem for extension files)
+
```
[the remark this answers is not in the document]
robshakir: Agreed, given that this is the case can we maybe add a short comment in this `README`? I don't think we need to commit to deprecation yet, but maybe signposting this is reasonable?

dplore:
> @hellt @rgwilton @earies -- PTAL, if you have any concerns please raise them here.
> 
> @dplore -- can we discuss this in the community meeting with a view to getting it merged?

 I have added to the July[ 10, 2025 OpenConfig Community Meeting agenda](https://docs.google.com/document/d/11sXUF2zujhbPs_bt-RFJhh2DcIbEvnOpd9GypE-u3cc/edit?tab=t.0).  @brianneville can present?

robshakir:
Thanks Brian.

There isn't a template -- it's fairly unstructured. You can use the actual code itself -- or just a brief overview of what the aim is, and what the rough design is would be sufficient.

Thanks again.

r.

extension/extension.proto
ncorran: Agree - its not a good name.  Most 'extensions' == bugfixes.  Some could be new functionality, but perhaps some other name would be generic enough but more clear on what is happeing, such as 'package' 'software-bundle' or ...

rgwilton: Hi Brian,

Are there other examples of software/package installation systems that use the term extension other than Arista EOS?  Extension is very generic term - if we are able to choose a more packaging specific name then that would likely to provide greater clarity in the longer term.

Also noting that fix a bug isn't really extending the system at all, it is just installing a software package to fix a bug.  I also don't see that upgrading the software is an extension.

brianneville: Maybe something like `SoftwareBundle` (as mentioned in the comment above) would be a better name instead of `Extension`? 
I'm hesitant to use the terminology of `Package` because i dont want this service or its RPCs to collide with the `SetPackage` RPC or its messages which already use that term.

extension/extension.proto
```diff
@@ -0,0 +1,221 @@
...
+
+  // UninstallExtension uninstalls an extension.
+  // It returns an error if the requested extension is not installed.
+  rpc UninstallExtension(UninstallExtensionRequest)
+      returns (UninstallExtensionResponse);
+}
+
+message ListExtensionsRequest{
+}
+message ListExtensionsResponse{
```
iankimpton, line 88: Related to the "what is an extension" questions above:
I would expect that installing an 'extension' which is a tarball/zipfile of multiple RPMs to be supported. However, I would not expect the tarball/zipfile to ever appear here as 'installed' - it's just a convenient delivery mechanism for bundling multiple packages.

extension/extension.proto
```diff
@@ -0,0 +1,221 @@
...
+message FinalizeExtensionsResponse{
+}
+
+message UninstallExtensionRequest{
+  // name of the extension to be uninstalled
+  string name = 1;
+
+  // a boolean indicating whether we should force uninstalling this extension.
+  // if set to true, then ignore conflicts and dependencies when uninstalling.
+  bool force = 2;
```
iankimpton, line 155: Two comments on this:

1) I'd assume support for this is optional
2) Does it make more sense to be consistent with InstallExtensionRequest make sense? 'ignore_dependency_checks' vs 'force'

extension/extension.proto
```diff
@@ -0,0 +1,221 @@
...
+}
+
+message FinalizeExtensionsRequest{
+}
+message FinalizeExtensionsResponse{
+}
+
+message UninstallExtensionRequest{
+  // name of the extension to be uninstalled
+  string name = 1;
```
iankimpton, line 151: Where an extension has been installed to bugfix/resolve a CVE for example, then 'removing' it would generally mean some sort of downgrade operation. So in that case is this saying 'downgrade to the next highest version you have'?

So to pick a concrete example:
 * Let's say I'm working with RPMs and I bugfix openssh
 * But then I want to remove the bugfix
 * I think in RPM terms this would mean reverting to the earlier version of openssh
So in order for this to 'just work' one would need to have the lower version of the RPM available to install

extension/extension.proto
```diff
@@ -0,0 +1,221 @@
...
+    // the extension may be installed but no longer available
+    // if the extension file has been removed.
+    bool available = 9;
+
+    // any errors attached to this extension
+    string errors = 10;
+    repeated Package packages = 11;
+
+    // affected_services is a list of the services which can be affected by this
+    // extension. This is static information
```
iankimpton, line 130: >> // extension. This is static information

If this delivers a library, then knowing 'statically' (i.e. upfront) what is affected seems very difficult, especially for dynamically linked libraries - and the set of possible services can vary depending on the extensions running etc.

extension/extension.proto
```diff
@@ -0,0 +1,221 @@
...
+    repeated string affected_services = 12;
+
+    // services_to_restart is a list of the services which are pending a reload
+    // to be finalized.
+    // This list will be cleared after a ReloadExtended RPC executes.
```
iankimpton, line 135: I'm not sure what a ReloadExtended RPC is, but I might be missing something? gnoi system.Reload?

extension/extension.proto
```diff
@@ -0,0 +1,221 @@
...
+    // UNSPECIFIED should be treated as an error.
+    UNSPECIFIED = 0;
+    INSTALLED = 1;
+    NOT_INSTALLED = 2;
+
+    // This extension was force-installed - checks regarding
+    // signature or dependencies may have been ignored
+    FORCE_INSTALLED = 3;
+  }
+  // an extension may contain multiple packages
```
iankimpton, line 99: In the case where an extension 'is' a single package, say an RPM, does this field repeat the name/version/release/etc of an extension, or would we expect the package list to be empty?

extension/extension.proto
```diff
@@ -0,0 +1,221 @@
...
+  // for an extension which is already present on the system
+  string extension_name = 1;
+
+  // if this is true, then the system will also
+  // ensure that this extension is loaded on boot/reboot
+  bool load_after_boot = 3;
+
+  // if this is true, then the system will ignore
+  // any errors which could arise from missing dependencies
+  bool ignore_dependency_checks = 4;
```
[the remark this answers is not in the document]
earies: And it imo effectively means the same thing (e.g. skip checks) - lets at least use consistent terminology across operations

extension/extension.proto
earies: At a minimum we should at least consider a common definition that is reused/import vs. recreating the same or similar messages structs if we want to contain such operations within the context of specific APIs for abstraction purposes.

As for the reasons you list - sure you can get more granular on some authorization control and ease of use but I'd also expect operational workflows to potentially disassociate the action of file transfer from that of installation/activation - e.g. transfer and verify files out to the fleet for >= 1 purpose (pre-stage, reduce time), follow up w/ install/activation later.  For that workflow one could use either the generic API or one that bakes in their own specific file xfer method.

extension/extension.proto
earies: Agreed - the "input" to extension operations here can very well unfold into multiple extensions of varying names and versions to be installed - vendor provided, user built, etc... In fact in such scenarios it also cannot be guaranteed that the encapsulated packaging of a collection have a meaningful name or version to itself that is retained or related.

Would you mind elaborating what some example `.txtpb` might look like for say a situation of a .swix or other bundle that encapsulates multiple underlying packages?

earies (commented):
Just getting around to this as well and did an initial pass - I've also forwarded to a few additional folks on our side for further review

Few high level comments:
* As was already noted above, in our implementation we use the same underlying operation whether it is extensions or a base OS image itself
* It would be good if you can provide some example `.txtpb` of these messages in accordance to operations and situations (e.g. single package, package that encapsulates other packages)
* If we are going to replicate behaviors across various APIs for containment purposes, should we consider at least reuse of message fields from common definitions vs. reinvention and possible divergence

extension/extension.proto
brianneville: renamed to SoftwareBundle in https://github.com/openconfig/gnoi/pull/160/commits/7625e294936f3f276924939be70b11cc29a21af3

extension/extension.proto
```diff
@@ -0,0 +1,221 @@
...
+
+    // services_to_restart is a list of the services which are pending a reload
+    // to be finalized.
+    // This list will be cleared after a ReloadExtended RPC executes.
+    repeated string services_to_restart = 13;
+  }
+  repeated Extension extensions = 1;
+
+  // extension_directory is a directory where extension files are contained.
+  string extension_directory = 2;
```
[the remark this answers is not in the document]
brianneville: removed in https://github.com/openconfig/gnoi/pull/160/commits/a76ec7ef486af6087df9f57a23b45b83b8736b91

extension/extension.proto
```diff
@@ -0,0 +1,221 @@
...
+// The extension is applied to the running configuration.
+message InstallRequest{
+  // extension_name is the name of the extension to install.
+  // This is required because InstallExtension can be called
+  // for an extension which is already present on the system
+  string extension_name = 1;
+
+  // if this is true, then the system will also
+  // ensure that this extension is loaded on boot/reboot
+  bool load_after_boot = 3;
```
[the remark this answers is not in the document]
brianneville: This field about having the software bundle be able to be applied if the system is rebooted. 
Typical reboots will should reboot the base OS (which doesnt include applying the software bundles). 
This field allows for clients to decide if they want this software bundle to be re-applied when the system reboots.
Essentially it can be seen as marking the software bundle as "persistent". 
I'll push a commit updating the comment and renaming both this `load_after_boot ` var and the `install_at_boot` var to `persistent` for consistency

extension/extension.proto
```diff
@@ -0,0 +1,221 @@
...
+    // installed on next system reload
+    bool install_at_boot = 7;
+
+    // bool indicating if this extension has a valid signature
+    bool signature_valid = 8;
+
+    // available is true if the extension file is still available
+    // on the system.
+    // the extension may be installed but no longer available
+    // if the extension file has been removed.
```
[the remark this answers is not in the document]
brianneville: yep, the underlying implementation could remove it, or someone could ssh onto the system and remove it manually if they like.
If an software bundle was uninstalled and then the system didnt have the source for it (i.e. it was unavailable), then the system wouldnt be able to reinstall it from source, and would need the software bundle to be transferred again

extension/extension.proto
```diff
@@ -0,0 +1,221 @@
...
+  message Extension {
+    string name = 1;
+    string version = 2;
+    string release = 3;
+    string description = 4;
+    InstallStatus install_status = 6;
+
+    // install_at_boot is true if the extension is set to be
+    // installed on next system reload
+    bool install_at_boot = 7;
```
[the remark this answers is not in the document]
brianneville: This is indicating whether the changes made by the software bundle will be able to survive a system reboot. see the comment on https://github.com/openconfig/gnoi/pull/160#discussion_r2223491704.

In https://github.com/openconfig/gnoi/pull/160/commits/6a57f982ac5acabc24237c5475727464fb4a2499 i've renamed this to "persistent" to better clarify what this means.

extension/extension.proto
brianneville: I'd expect this kind of workflow to be fairly uncommon, but i think if a network operator wanted to rollback their system like this they could just reboot the system, and this would cause the system to startup with:
- the base OS and whatever RPMs it had initially
- any software bundles which were marked as `persistent` (obviously this wouldnt include uninstalled software bundles)

For "what happens directly after uninstallation" i'd say we could just leave it up to the system. 
If its possible to rollback like that then system should try to do so, but it may not be possible to rollback (e.g. if some previous dependancy files were overwritten).

extension/extension.proto
```diff
@@ -0,0 +1,221 @@
...
+  // UninstallExtension uninstalls an extension.
+  // It returns an error if the requested extension is not installed.
+  rpc UninstallExtension(UninstallExtensionRequest)
+      returns (UninstallExtensionResponse);
+}
+
+message ListExtensionsRequest{
+}
+message ListExtensionsResponse{
+  enum InstallStatus {
```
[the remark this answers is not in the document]
brianneville: The OS service in os.proto is only for transferring bootable images onto the system - these smaller software bundles are not really bootable images - we're not transferring a whole OS here  :)

The OS image doesnt show up here. information about the current OS image should be found by running the OS/Verify RPC:
https://github.com/openconfig/gnoi/blob/d04468fba8d73905e7865fcfd0076a48523f71e6/os/os.proto#L147-L156

extension/extension.proto
brianneville: I'm thinking we can just remove this `packages` field and the `Package` message type and just not expose the inner contents of the software bundle. 

This RPC to list the software bundles should operate on the same granularity of the other RPCs. The gnoi client deals with only the software bundle as a unit that it installs/uninstalls, so the RPC listing the software bundles should do the same. 

>However, I would not expect the tarball/zipfile to ever appear here as 'installed' - it's just a convenient delivery mechanism for bundling multiple packages.

The InstallSoftwareBundle/UninstallSoftwareBundle RPCs can only operate at the level of the software bundle, so i think it makes sense to say "this whole bundle is installed" or "this whole bundle is uninstalled". It may be the case that a software bundle has unfolded into multiple packages, but the client has no control over these individual packages. 
- If the software bundle was something which was provided by the vendor then this lack of fine-grained control is ok, because the software bundle represents some singular feature change/bugfix, which is either installed or not.
- If the software bundle was something which was created by the network operator then this lack of fine-grained control is also ok, because the network operator themselves can decide the contents of the bundle (as well as the name/description of it). For example, if the network operator wanted to install 10 RPMs and be able to control all of them individually, they could just run the InstallSoftwareBundle RPC 10 times, one for each RPM. This is less convenient sure, but it would work.

extension/extension.proto
brianneville: I would expect that it repeats the name/version/release of the RPM

(although im thinking that we should remove this `Package` message, see my comment above: https://github.com/openconfig/gnoi/pull/160#discussion_r2225350372 )

extension/extension.proto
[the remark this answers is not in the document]
brianneville: this should have been referring to system processes, apologies for the confusion.
I've updated this in https://github.com/openconfig/gnoi/pull/160/commits/11727981a447d082bb67bd4f0d7ac5ed546bbfaa

brianneville: ack, yeah thats fair. I've updated this comment to note that this list may be incomplete, see
```
    // affected_processes is a list of the processes which can be affected by this
    // software bundle. This is static information and may be incomplete (e.g. in
    // in the case of dynamically linked libraries).
```

https://github.com/brianneville/gnoi/blob/11727981a447d082bb67bd4f0d7ac5ed546bbfaa/software_bundle/software_bundle.proto#L139-L140

extension/extension.proto
brianneville: bumping this thread - do you guys have any further thoughts or concerns about the above proposal to remove the `packages` field?
If not i can go ahead and update the PR

software_bundle/software_bundle.proto
```diff
@@ -0,0 +1,229 @@
...
+    // Creating and populating SSL profiles is out of the scope of this
+    // service and should be done through some other means (e.g. gNSI Certz).
+    string ssl_profile_id = 6;
+  }
+}
+
+message TransferRequest{
+  oneof request {
+    // name of the software bundle
+    string software_bundle_name = 1;
```
iankimpton, line 223: Some system support the ability to install a tarball of software bundles (i.e. several vendor-specific software bundles get concatenated into a tar/cpio/zip archive), which can be applied at all once as a convenience message (rather than copied and then installed separately)

Is it valid for the contents of this TransferRequest to be such a set of files grouped together?

brianneville: >Is it valid for the contents of this TransferRequest to be such a set of files grouped together?

Sure, yep it would be valid - the software bundle is an arbitrary blob which is understood by the system. As long as the system knows how to unpack this tar/cpio/zip archive, then you can send it over.

software_bundle/software_bundle.proto
```diff
@@ -0,0 +1,229 @@
...
+  rpc ListSoftwareBundles(ListSoftwareBundlesRequest)
+      returns (ListSoftwareBundlesResponse);
+
+  // FinalizeSoftwareBundles reloads any services on the system for
+  // which software bundles have been installed but which have not yet been
+  // reloaded with those software bundles.
+  // Please note that this may kill the connection if the service
+  // hosting the gRPC server reloads so the server is not guaranteed
+  // to return a response here.
+  rpc FinalizeSoftwareBundles(FinalizeSoftwareBundlesRequest)
```
dplore, line 30: Will this reload to restart processes?  Is there a need to differentiate?  Operationally I might want to know if the "finalize" function is service impacting or not.

software_bundle/software_bundle.proto
```diff
@@ -0,0 +1,229 @@
...
+  // STREAM END
+  //
+  // 3. To install a pre-existing software bundle:
+  // client ______________________________________ server
+  // InstallRequest (with software_bundle_name set)
+  // (note, other fields may be set )          -->
+  // .. server installs software bundle according to fields.
+  //                                           <-- InstallSoftwareBundleResponse
+  //                              (with acknowledgement = INSTALL_COMPLETE)
+  // STREAM END
```
dplore, line 73: perhaps move to a sequence diagram placed in the README.

software_bundle/software_bundle.proto
```diff
@@ -0,0 +1,229 @@
...
+    // by this software bundle.
+    // This list may be incomplete if the system cannot determine all of the
+    // processes that could be affected.
+    repeated string affected_processes = 9;
+
+    // processes_pending_restart is a list of the processes which are waiting
+    // be restarted to apply the software bundle.
+    // This list will be cleared after a FinalizeSoftwareBundles RPC executes
+    // and restarts the affected processes.
+    repeated string processes_pending_restart = 10;
```
dplore, line 141: How do we know the operational state of the processes, related to the software bundles?      

If the process isn't pending restart, does that mean the installed bundle is active?   

maybe a direct indication would be helpful.  Like
```
  enum ActivationStatus {
    // UNSPECIFIED should be treated as an error.
    UNSPECIFIED = 0;
    NOT_ACTIVATED = 1;
    ACTIVATED = 2;
  }
```

dplore (commented):
Added a few suggestions.

dplore:
@brianneville ,  also, please provide the following in your README:
- Why not use or extend gnoi.OS?
- State the network operator workflow and/or use cases.  eg: why does the operator want to install a package?  Why do they want to have this level of visibilty and control?  (Why not upgrade the OS?  Why not have an opaque OS level patch vs. needing to see the list of affected processes and wether or not they have restarted, etc?).  Feel free to link to/leverage existing documentation as evidence that these services are in fact already implemented and being used.

dplore:
In review at OC Operators meeting 2025-11-04, there's a question about how to manage storage space?  Is there any way to remove a package that's been transferred?  How is storage space cleaned up?

brianneville:
>In review at OC Operators meeting 2025-11-04, there's a question about how to manage storage space? Is there any way to remove a package that's been transferred? How is storage space cleaned up?

Currently the `UninstallSoftwareBundle` RPC description notes:
```
  // UninstallSoftwareBundle uninstalls an software bundle.
  // It returns an error if the requested software bundle is not installed.
  rpc UninstallSoftwareBundle(UninstallSoftwareBundleRequest)
      returns (UninstallSoftwareBundleResponse);
```
But we could change it to note that the the RPC is idempotent so it doesnt return an error, and then add a field like `cleanup_bundle` in the `UninstallSoftwareBundleRequest` that would instruct the device to cleanup the software bundle from the filesystem.

This would allow the operator to run one `UninstallSoftwareBundle` RPC to uninstall the software bundle, and then later they can run the `UninstallSoftwareBundle` again with the `cleanup_bundle` field set to cleanup the storage.
Or they can run just one `UninstallSoftwareBundle` RPC with the `cleanup_bundle` field set, to both uninstall and cleanup in a single RPC.

software_bundle/software_bundle.proto
```diff
@@ -0,0 +1,193 @@
...
+  rpc ListSoftwareBundles(ListSoftwareBundlesRequest)
+      returns (ListSoftwareBundlesResponse);
+
+  // FinalizeSoftwareBundles reloads any processes on the system for
+  // which software bundles have been installed but which have not yet been
+  // reloaded with those software bundles.
+  // Please note that this may kill the connection if the process
+  // hosting the gRPC server reloads, so the server is not guaranteed
+  // to return a response here.
+  rpc FinalizeSoftwareBundles(FinalizeSoftwareBundlesRequest)
```
dplore: Minor suggestion:

```suggestion
  rpc ActivateSoftwareBundles(FinalizeSoftwareBundlesRequest)
```

software_bundle/README.md
```diff
@@ -0,0 +1,219 @@
...
+#### What is a software bundle?
+A software bundle is a single file (e.g. RPM, squashfs, zip) which can be installed
+on the system to extends the functionality of the system in some way.   
+A software bundle may be installed to:
+- fix some bug/CVE on the system, by updating the programs affected by that
+  bug/CVE
+- add a new feature to the system
+- add/update some software utility (e.g. perl, ncurses)
+- update some dependency (e.g. updating some python site-package)
+
```
dplore: The README should clarify if multiple versions of a software bundle may exist on the system (presumably only one is activated at any time).  Install/Uninstall need to be updated to handle picking which version is being used.

software_bundle/README.md
```diff
@@ -0,0 +1,219 @@
...
+#### Comparison with the gNOI OS package:
```
dplore: formatting nit:  Can you create a header for "Comparisions with other gNOI RPCs" near the end of this README and then move the OS and SetPackage comparisons into it?

brianneville:
updated to address comments. 

> Also, should List, Install and UninstallSoftwareBundleRequest include a version field? Or is the intent that there must only be one copy of a named SoftwareBundle on the device? If there is an activated Bundle and one which is not activated, how is that listed?

to keep things simple, we can just insist that the name is unique. 
The name is free-form, so operators can just rename bundles whatever they want to avoid collisions .
I also updated the proto to note that we should return an ALREADY_EXISTS error if a non-unique name is provided.

operators wishing to install something with a conflicting name should just deactivate/remove the already-existing bundle.

> The data returned by the List RPC should probably also have OC Yang paths which mirror the ListResponse data.

This can be addressed in the openconfig/public repo if desire is there, but it would be out of scope of this PR

software_bundle/software_bundle.proto
brianneville: this RPC was renamed to `Activate` in the latest patch set

software_bundle/README.md
brianneville: done in https://github.com/openconfig/gnoi/pull/160/commits/4df2a72eaa307b8fbead852292a1f1293b34491a and included the suggestion that operators append the version into the name.

software_bundle/README.md
brianneville: done in https://github.com/openconfig/gnoi/pull/160/commits/4df2a72eaa307b8fbead852292a1f1293b34491a

PR #162, opened 2024-02-12
Adding support for volumes in containerz

alshabib:


containerz/containerz.proto
```diff
@@ -208,13 +221,25 @@ message ListResponse {
   Status status = 4;
 }
 
-message StartRequest {
+// A volume can be passed to a container such that it can persist data.
```
robshakir, line 224: Should we have more definition here around what a volume is expected to do? I think we want to say that it's backed by persistent storage and must exist across device reboots right?

containerz/containerz.proto
```diff
@@ -208,13 +221,25 @@ message ListResponse {
...
+message Volume {
+  string name = 1;
+
+  // The mount_point is location with the container where the volume should be
+  // attached.
```
robshakir, line 229:
```suggestion
  // attached within the container.
```

containerz/containerz.proto
```diff
@@ -299,3 +327,69 @@ message LogRequest {
 message LogResponse {
   string msg = 1;
 }
+
+message CreateVolumeRequest {
+  // name of the volume. If present, this system must create the volume with
+  // this name otherwise it must allocate a name.
+  string name = 1;
+
+  // The volume driver. If not specified, this defaults to the the standard
```
robshakir, line 336: Can this be an enumerated value? It seems like here we're very flexible, and there's maybe not a clear definition that will work across different implementations. It's also not clear how to express the minimum that `Containerz` is looking for -- e.g., is anything other than `MOUNT` actually required?

containerz/containerz.proto
```diff
@@ -299,3 +327,69 @@ message LogRequest {
...
+  // linux mount operation (http://man7.org/linux/man-pages/man8/mount.8.html)
+  // and associated options.
+  string driver = 2;
+
+  // Options for the driver. The actual option keys and values are driver
```
robshakir, line 341: Can you give some examples here? Are you thinking about e.g., flags to `mount` that are supplied? I would prefer less "open" modelling here, so that you have something like:

```protobuf
enum Driver { 
  DS_UNSPECIFIED = 0;
  DS_MOUNT = 1;
  DS_NFS = 2;
  ...
}

message MountOptions {
  bool read_only = 1;
}

message NFSOptions {
  uint32 reconnect_time_seconds = 1;
}

Driver driver = 1;

oneof options {
  MountOptions mount = 2;
  NFSOptions nfs = 3;
  ...
}
```

where now we have much more explicitly specified options that might be implemented. I understand that this might need changes to `Containerz` when we're adding new drivers, but my concern is whether the very generic API we have here maps cleanly across different container runtimes. (Like, does this work for `hostPath` in k8s, or `configMap`?)

containerz/containerz.proto
```diff
@@ -299,3 +327,69 @@ message LogRequest {
...
+  // specific.
+  map<string, string> options = 3;
+
+  // Labels to apply to the volume. Labels are metadata for the the volume.
```
robshakir, line 345: Can you explain what the use case for this is?

containerz/containerz.proto
```diff
@@ -299,3 +327,69 @@ message LogRequest {
...
+}
+
+message ListVolumeRequest {
+  // List all volumes.
+  bool all = 1;
+
+  // Only return limit number of volumes.
+  int32 limit = 2;
+
+  // A filter is a mapping from a key to a list of values.
```
robshakir, line 369: Is this a requirement today (and do you expect different backing container runtimes to support this same filtering)?

containerz/containerz.proto
alshabib: done - I added this to the CreateVolume RPC since this message is only intended to attach an existing volume to a container.

containerz/containerz.proto
alshabib: This allows to label/tag volumes with information about what they contain or even  #perhaps version information (e.g. tagging the config version allowing rollbacks to another version more easily) which can then be filtered.

containerz/containerz.proto
alshabib: See comment above

containerz/containerz.proto
alshabib: For example

```
CreateVolumeRequest {
   name = "volume"
   driver = "local"
   options = {
      type = none
      o = bind
      device = /some/host/path
   }
}
```

Would result in a volume backed by the host path `/some/host/path` with the name `volume`.

I think this would work pretty well across across runtimes as all the details are passed to the runtime and the driver. If we are more explicit we would need to map that logic in the containerz server which could lead to large implementation.

compile_protos.sh
```diff
@@ -4,7 +4,7 @@ set -euo pipefail
 proto_imports=".:${GOPATH}/src/github.com/google/protobuf/src:${GOPATH}/src/github.com/googleapis/googleapis:${GOPATH}/src:."
 
 # Go
-for p in types bootconfig common containerz debug diag bgp cert factory_reset file healthz layer2 mpls system os otdr wavelength_router packet_link_qualification; do
+for p in types common containerz debug diag bgp cert factory_reset file healthz layer2 mpls system os otdr wavelength_router packet_link_qualification; do
```
robshakir, line 7: Re-add `bootconfig` here.

containerz/containerz.proto
```diff
@@ -299,3 +328,92 @@ message LogRequest {
 message LogResponse {
   string msg = 1;
 }
+
+enum Driver {
```
robshakir, line 332: Suggest to add a comment to explain what `Driver` is (since its a volume driver, but top-level description is here).

robshakir (approved):

PR #163, opened 2024-02-13
Update bootconfig.proto

sachendras:
Update bootconfig.proto to include updating security artifacts aswell part of the SetBootConfigRequest message.

### DIFF
--- a/bootconfig/BUILD.bazel
+++ b/bootconfig/BUILD.bazel
@@ -28,6 +28,10 @@ proto_library(
     deps = [
         "//types:types_proto",
         "@com_github_openconfig_bootz//proto:bootz_proto",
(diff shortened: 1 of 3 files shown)


marcushines (commented):
need to fix build but otherwise lgtm

xw-g:
gNSI.x client has access to those gNSI namespaces. I don't think we should add those in this API (which is about Bootconfig namespace)

morrowc:
I think the BUILD file and WORKSPACE are asking for things that one doesn't have.. this seems to move things along a bit more:

  `bazel run //:gazelle -- update-repos github.com/openconfig/gnoi/types`

NOTE: I have no idea what that does... but... not erroring as it was before, for me.

PR #188, opened 2024-04-24
proto updates for PROBE (RFC8335)

fenner:
This is a new RPC to request a remote PROBE session.  This is heavily based on PingRequest/PingResponse, modified for the PROBE semantics ( https://datatracker.ietf.org/doc/html/rfc8335 ).

JunOS and Arista EOS have their own Probe clients; there's a WIP at https://github.com/aristanetworks/iputils/tree/probe-client to make the iputils "ping" client able to send PROBEs.


system/system.proto
```diff
@@ -227,6 +232,86 @@ message PingResponse {
...
+  Active active = 6;
+
+  // summary
+  int32 sent = 7;       // Total packets sent.
+  int32 errors = 8;     // Count of errors received when sending packets.
+  int32 received = 9;   // Total packets received.
+  // histograms in summary
+  map <uint32,uint32> codes = 10;     // Keys are values of Code enum
+  map <uint32,uint32> states = 11;    // Keys are values of State enum
+  map <uint32,uint32> actives = 12;   // Keys are values of Active enum
```
LimeHat, line 312: Is this really needed?
gNOI client can do the math, adding a summary to each response seems unnecessary

system/system.proto
```diff
@@ -227,6 +232,86 @@ message PingResponse {
...
+    STATE_FAILED = 6;
+  }
+  enum Active {
+    ACTIVE_UNKNOWN = 0;
+    ACTIVE_INACTIVE = 1;
+    ACTIVE_NOV4_NOV6 = 2;
+    ACTIVE_V4 = 3;
+    ACTIVE_V6 = 4;
+    ACTIVE_V4_V6 = 5;
+  }
```
LimeHat, line 291: I would suggest mapping the v4/v6 bits directly to individual fields, like it's done in the RFC.

system/system.proto
```diff
@@ -227,6 +232,86 @@ message PingResponse {
...
+  string source = 1;
+  int64 time = 2;
+
+  // For individual responses
+  int32 sequence = 3;
+  Code code = 4;
+  // state is only present if code is CODE_NO_ERROR
+  // and request was remote
```
LimeHat, line 299: proto3 does not allow to distinguish between a non-present enum and the enum with the default value ("0").

You can replace "present" with "non-zero", or remove this altogether (since you're following the rfc semantics)

system/system.proto
```diff
@@ -227,6 +232,86 @@ message PingResponse {
...
+  enum Code {
+    CODE_UNKNOWN = 0;
+    CODE_NO_ERROR = 1;
+    CODE_MALFORMED_QUERY = 2;
+    CODE_NO_SUCH_INTERFACE = 3;
+    CODE_NO_SUCH_TABLE_ENTRY = 4;
+    CODE_MULTIPLE_INTERFACES = 5;
+  }
+  enum State {
+    STATE_UNKNOWN = 0;
```
LimeHat, line 276: `STATE_RESERVED` to follow the rfc

system/system.proto
```diff
@@ -227,6 +232,86 @@ message PingResponse {
...
+  string source = 1;
+  int64 time = 2;
+
+  // For individual responses
+  int32 sequence = 3;
+  Code code = 4;
+  // state is only present if code is CODE_NO_ERROR
+  // and request was remote
+  State state = 5;
+  // active is only present if code is CODE_NO_ERROR
```
LimeHat, line 301: same as above

LimeHat (changes requested):

system/system.proto
fenner: That's true - other than the errors, all of these values can be derived from the set of responses. I copied this concept from the "ping" response, which also includes summary values that can be derived from the set of responses. If it's useful there, then I thought it would be useful here.

system/system.proto
fenner: This structure was used based on the section A.1 that was added to the draft RFC update based on user feedback.

https://www.ietf.org/archive/id/draft-fenner-intarea-probe-clarification-00.html#name-information-display

While I think it's possible to create a tool that performs the GNOI transaction and presents the information to the user, I've seen users run the RPCs and just want to be able to interpret the raw json results, so thought it was better for the json results to be in a presentation format.

system/system.proto
robshakir: I think its fine to have this as `UNKNOWN`. If we want to have a reserved value then we should probably have a different value that doesn't have semantics associated with it. proto style is for this to be `UNKNOWN`/`UNSPECIFIED` rather than have semantic meaning.

robshakir (commented):
Generally LGTM.

system/system.proto
LimeHat: > so thought it was better for the json results to be in a presentation format.

I'm not sure if the value `5` is more useful to the user than `active true, ipv4 true, ipv6 false`? :-)

system/system.proto
LimeHat: Personally, I think the direct streaming of responses is a better approach, but if you want to provide an option to return the summary, I would suggest refactoring this to use the `oneof` model and splitting it into two message types.

system/system.proto
```diff
@@ -227,6 +232,86 @@ message PingResponse {
...
 }
 
+// A ProbeRequest describes the probe operation to perform.  The
+// destination field and exactly one of the probe_target fields
+// are required.  Any field not specified is set to a reasonable server
+// specified value.  Not all fields are supported by all vendors.
+message ProbeRequest {
+  string destination = 1;   // Proxy address to probe. required.
+  string source = 2;        // Source address to probe from.
+  oneof probe_target {
```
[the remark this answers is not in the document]
fenner: Well, https://fenner.github.io/probe-clarification/draft-fenner-intarea-probe-clarification.html refers to interface name and ifindex from https://www.rfc-editor.org/rfc/rfc8343 - I'm pretty ignorant of the details of ietf netmod vs. openconfig models.  I'd be happy to put in appropriate references, though, if you can help me find them.

ifaddr is expected to be the IP address assigned to some interface (presumably a state lookup, not a config one).

remote is expected to be an entry in the ARP or ND table.

The reference is also a little tricky because these are not values on the node that the RPC is being sent to, but instead, values on the `destination` that's being specified in the RPC.

system/system.proto
fenner: I changed this to "state is only *valid* if code is CODE_NO_ERROR", I think that sounds right.

system/system.proto
fenner: Same as above, used *valid*.

system/system.proto
fenner: I assumed that the gnmi client would translate enum values, so you would get ACTIVE_V4 - but I admit ignorance on this front.

Another reason to use this enum is for the histogram summary - a histogram of "5 active, 3 ipv4, 1 ipv6" doesn't really tell you whether ipv4+ipv6 were ever active together.  (I know you don't see the point of the summary data, so this won't be a particularly convincing argument).

system/system.proto
fenner: As I said, I modeled this after "ping" to keep it similar since the protocols are similar. Should "ping" also use `oneof`?

system/system.proto
fenner: Thanks, Rob!

system/system.proto
LimeHat: >Should "ping" also use oneof?

Maybe :-) But it's a bit too late to change that, and not the point. 

In this case, the state machine is significantly more complex than in ping rpc. IMO it's easier to write a client that doesn't have to infer the meaning of the message (a summary vs an individual response) based on field values, especially considering that you mostly have independent fields in each case. That also automatically answers questions such as "what fields are expected or allowed to be set together, what are valid/invalid combinations". 

Is there any concern about this?

fenner: I can make the change, I'm just concerned about creating a difference from the ping RPC.  I figured that the ping definition established how this kind of RPC was desired to be expressed in GNOI.  (For example, I used an int64 for the time, even though uint64 is a better fit for an absolute timestamp, because that's what ping did).

system/system.proto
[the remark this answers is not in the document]
fenner: @robshakir just to clarify what you're looking for here: do you want comments like

```
message ProbeRequest {
  string destination = 1;   // Proxy address to probe. required.
  string source = 2;        // Source address to probe from.
  oneof probe_target {
  oneof probe_target {
    // exactly one of the next 4 items must be present.
    // ifIndex to probe, if present.  Equal to the value of
    // /interfaces/interface/state/if-index for the requested
    // interface on the destination node.
    int32 ifindex = 3; 
    // ifName to probe, if present.  Equal to the value of
    // /interfaces/interface/state/name for the requested
    // interface on the destination node.
    string ifname = 4;
    // ifName to probe, if present.  Equal to the value of
    // /interfaces/interface/<... IP address assigned ...> for the requested
    // interface on the destination node.
    string ifaddr = 5;
    // Remote address to probe, if present.  Equal to the value of
    // /neighbors/something/<... ARP or ND table entry ...> for the requested
    // remote neighbor on the destination node.
    string remote = 6;
  }
```

where obviously I'd fill in the right paths for "interface IP address" and "ARP/ND table" before publishing a revision

PR #200, opened 2024-06-04
Update RPC and other start parameters.

alshabib:
This commit adds an UpdateContainer RPC as well as some parameters for starting containers.


containerz/containerz.proto
```diff
@@ -64,6 +80,22 @@ service Containerz {
   // (i.e. kill) a container.
   rpc StopContainer(StopContainerRequest) returns (StopContainerResponse) {}
 
+  // UpdateContainer updates a running container to the image specified in the
```
LimeHat, line 83: Can you expand on the meaning of this?

Is it fair to say that the combined operation is a rough equivalent to "stop the container, delete the container, launch the container with the new parameters"?

If the container exists but not running at the time the request was received, is it considered an error?

containerz/containerz.proto
```diff
@@ -64,6 +80,22 @@ service Containerz {
...
+  // request. By default the operation is synchronous which means that the
+  // request will only return once the container has either been successfully
+  // updated or the update has failed. If the client requests an asynchronous
+  // update then the server must perform all validations (e.g. does the
+  // requested image exist on the system or does the instance name exist) and
+  // return to the client and the update happens asynchronously. If the update
+  // fails, the server must restore the previous version of the container. It
```
LimeHat, line 90: > the previous version of the container

The previous container is likely to be deleted; it is possible to restore the previous image / launch params, but the container is likely to be re-created. The end result is likely to be identical in most cases, but non-persistent data might be lost (if any).

One potential downside of doing this type of combined operation on the server side is that the rollback result will be invisible to the caller: it will be notified that the new deployment failed, but it won't get a notification whether the server was able to launch the previous version or not.

alshabib: That's correct the previous container instance will most likely be gone and any data within it will be lost unless the data was written to a volume. That volume can be reattached when the upgrade terminates. Generally storing data within the container is considered an anti pattern for exactly this reason.

 Yes in async mode the client would need to check the result of the update but otherwise the result of the upgrade would be communicated back to client.

containerz/containerz.proto
alshabib: Good question - yes this is equivalent and could be implemented with existing RPCs. We introduced this RPC to formalize the contract with the user.

Yes this should be an error - I will update comment. Thanks for pointing that out.

containerz/containerz.proto
LimeHat: >Yes in async mode the client would need to check the result of the update but otherwise the result of the upgrade would be communicated back to client.

Result of the upgrade - yes, but if the upgrade failed, the last operation ("restoring the previous version") will not produce a result that can be communicated back to the client. I see that you added an additional error code to the latest revision:
```
   // The update failed and the previous version was restored.
    NOT_UPDATED = 4;
```

Which, presumably, means that now the result will be returned only after the rollback is done?

Note that another possible outcome is that the update failed and the previous version was _not_ restored; not sure if you want to add another explicit code for that.

containerz/containerz.proto
```diff
@@ -37,18 +37,34 @@ option (types.gnoi_version) = "0.1.2";
 
 service Containerz {
   // Deploy sets a container image on the target. The container is sent as
-  // a sequential stream of messages containing up to 64KB of data. Upon
-  // reception of a valid container, the target must load it into its registry.
-  // Whether the registry is local or remote is target and deployment specific.
-  // A valid container is one that has passed its checksum.
+  // a sequential stream of messages containing up to 4MB of data (max grpc
```
LimeHat, line 40: note that 4MB is the default max. value, but it is configurable. hence, this phrasing is ambiguous.

containerz/containerz.proto
alshabib: That's correct, in synchronous mode if the update fails the rpc should only return after the rollback is done.

I added a error code for a generic failure.

containerz/containerz.proto
alshabib: Thanks I mentioned that this is the default size.

containerz/containerz.proto
```diff
@@ -64,6 +80,24 @@ service Containerz {
...
+  // requested image exist on the system or does the instance name exist) and
+  // return to the client and the update happens asynchronously. If the update
+  // fails, the server must restore the previous version of the container. It
+  // is up to the client to check if the update actually updates the container
+  // to the requested version or not.
+  // If no StartContainerRequest is provided, the server must reuse the same
+  // run specification as the existing running containers. Otherwise, it must
+  // use the provided StartContainerRequest message.
+  // If a container exists but is not running then this operation should return
+  // an error.
```
kidiyoor, line 97: If a container is failing due to vendor OS upgrade i.e. incompatibility b/w application and OS, the best way to fix this is by rolling out the fix forward. So we should allow the upgrade to happen even if the old container is not running.

alshabib: Ahh good point - I changed it such that an upgrade should always happen.

kidiyoor (approved):

containerz/containerz.proto
```diff
@@ -37,18 +37,34 @@ option (types.gnoi_version) = "0.1.2";
 
 service Containerz {
   // Deploy sets a container image on the target. The container is sent as
-  // a sequential stream of messages containing up to 64KB of data. Upon
-  // reception of a valid container, the target must load it into its registry.
-  // Whether the registry is local or remote is target and deployment specific.
-  // A valid container is one that has passed its checksum.
+  // a sequential stream of messages containing up to 4MB of data (max default
+  // grpc message size). Upon reception of a valid container, the target must
```
robshakir, line 41:
```suggestion
  // gRPC message size). Upon reception of a valid container, the target must
```

containerz/containerz.proto
```diff
@@ -64,6 +80,24 @@ service Containerz {
...
   rpc StopContainer(StopContainerRequest) returns (StopContainerResponse) {}
 
+  // UpdateContainer updates a running container to the image specified in the
+  // request. By default the operation is synchronous which means that the
+  // request will only return once the container has either been successfully
+  // updated or the update has failed. If the client requests an asynchronous
+  // update then the server must perform all validations (e.g. does the
+  // requested image exist on the system or does the instance name exist) and
+  // return to the client and the update happens asynchronously. If the update
+  // fails, the server must restore the previous version of the container. It
```
robshakir, line 90: It's not quite clear to me what the order of operations of this RPC is expected to me. Is it make-before-break or break-before-make? I assume that `UpdateContainer()` being called will mean that we shut down the existing instance and then start a new using the new image. I assume that this'll be required, since otherwise resources (e.g., TCP ports) that are needed by the new instance may still be bound to the old instance.

Rollback then (as Sergey says above) is really "start a new container using the old image and config", I presume? Can we clarify?

containerz/containerz.proto
```diff
@@ -64,6 +80,24 @@ service Containerz {
...
+  // is up to the client to check if the update actually updates the container
+  // to the requested version or not.
+  // If no StartContainerRequest is provided, the server must reuse the same
```
robshakir, line 93: Is there a reason not to just require the `StartContainerRequest`? Can we also clarify that this is the `params` field in the `UpdateContainerRequest` in this RPC?

containerz/containerz.proto
```diff
@@ -78,6 +112,8 @@ service Containerz {
 
   // ListVolume returns all volumes that match the spec defined in the filter.
   rpc ListVolume(ListVolumeRequest) returns (stream ListVolumeResponse) {}
+
```
robshakir, line 115: (nit) remove blank lines.

containerz/containerz.proto
```diff
@@ -149,6 +185,37 @@ message ImageTransferSuccess {
...
+  // If force is set, the target should attempt to forcibly remove the
+  // container.
+  bool force = 3;
+}
+
+message RemoveImageResponse {
+  enum Code {
+    // An unspecified error. Must use the detail value to describe the issue.
+    UNSPECIFIED = 0;
```
robshakir, line 203: From https://protobuf.dev/programming-guides/dos-donts/#unspecified-enum -- we should generally avoid assigning semantics to the `_UNSPECIFIED`/0 value. Probably you want `INVALID` and `UNKNOWN` or similar here?

containerz/containerz.proto
```diff
@@ -257,6 +343,50 @@ message StartContainerRequest {
   // List of volumes that should be attached to the container.
   repeated Volume volumes = 7;
 
+  // The network to attach this container to. This could be 'host', 'bridged',
+  // or any other network available in the runtime.
+  string network = 8;
```
robshakir, line 348: Should this be an enumerated field, or is this entirely opaque to containerz?

containerz/containerz.proto
```diff
@@ -257,6 +343,50 @@ message StartContainerRequest {
...
+
+  message Capabilities {
```
robshakir, line 350: Can you document what capabilities are being described here? I didn't find any docs here or in the README.

(I assume that this covers capabilities(7), but I'm not sure -- if so, can't this be an enum?)

containerz/containerz.proto
```diff
@@ -257,6 +343,50 @@ message StartContainerRequest {
...
+      // Restart for max attempts if the container fails.
+      ON_FAILURE = 3;
+    }
+
+    Policy policy = 1;
+
+    // Only applies for the ON_FAILURE policy.
+    uint32 attempts = 2;
+  }
+  Restart restart = 10;
```
robshakir, line 379: Add documentation string for this field -- I presume this is set only as a container is started and tells us how to deal with restarting on failure of the container, or does it also apply to restart of the host system?

containerz/containerz.proto
```diff
@@ -257,6 +343,50 @@ message StartContainerRequest {
...
+
+  message RunAs {
```
robshakir, line 381: Can we document what this means? What happens if these fields are unspecified? Does this flag grant any host system privileges?

containerz/containerz.proto
```diff
@@ -257,6 +343,50 @@ message StartContainerRequest {
   // List of volumes that should be attached to the container.
   repeated Volume volumes = 7;
 
+  // The network to attach this container to. This could be 'host', 'bridged',
+  // or any other network available in the runtime.
+  string network = 8;
+
+  message Capabilities {
+    repeated string add = 1;
+    repeated string remove = 2;
```
robshakir, line 352: What are the semantics of `add` and `remove` here? What are we adding and removing from? Is it possible to just be declarative as to what we want the end state to be upon starting this container?

containerz/containerz.proto
```diff
@@ -318,6 +448,62 @@ message StopContainerResponse {
...
+  // The running containers name.
+  string instance_name = 1;
+
+  // Indicates if this is an async update.
+  bool is_async = 2;
+}
+
+message UpdateError {
+  enum Code {
+    // An unspecified error. The details field should provide more information.
```
robshakir, line 484: Same comment re: semantics of `UNSPECIFIED`/0 value as above.

robshakir (commented):
Added a number of clarification questions for the changes here.

Do we propose to support these changes in the reference implementation?

alshabib:
> Added a number of clarification questions for the changes here.
> 
> Do we propose to support these changes in the reference implementation?

Yes we will implement this changes in the reference implementation.

containerz/containerz.proto
alshabib: I explicitely mentioned that this is break-before-make. 

For rollback, the server can either restart the stopped container or it can start a new container with the old image. Both would be equivalent. I added text making this explicit.

containerz/containerz.proto
alshabib: In some cases we want to start the container with exactly the same resources as the previous container. It seems easier to do this by having the server inspect the running container, copy values and update the container. That being said - I am happy to make this field required.

containerz/containerz.proto
alshabib: This is opaque to containerz and could be any string. While docker defines some network types, a user could deploy a network driver and have a different network type.

containerz/containerz.proto
alshabib: Thanks - done.

containerz/containerz.proto
alshabib: Each policy enum has a description - I also added a generic description for the restart field.

containerz/containerz.proto
alshabib: done - PTAL.

containerz/containerz.proto
alshabib: By default a container starts with the following capabilities:
AUDIT_WRITE, CHOWN, DAC_OVERRIDE, FOWNER, FSETID, KILL, MKNOD, NET_BIND_SERVICE, NET_RAW, SETFCAP, SETGID, SETPCAP, SETUID, SYS_CHROOT.

Adding capabilities adds to this list while removing removes capabilities. Most applications will be fine with the default set and therefore this field can mostly be empty.

containerz/containerz.proto
robshakir: As per discussion -- the client should only depend on the process being restarted -- the contents of the filesystem are NOT guaranteed to be restored after a rollback operation.

containerz/containerz.proto
robshakir: Per discussion -- I'd prefer this is declarative so that we don't need to find the place where the container was created.

containerz/containerz.proto
robshakir: Can we add a comment that says that this is the default set?

robshakir: per discussion -- clarify what happens if we do `add = CAP_SYSADMIN`, `remove = CAP_SYSTEM`.

containerz/containerz.proto
robshakir: SGTM.

containerz/containerz.proto
```diff
@@ -257,6 +351,66 @@ message StartContainerRequest {
...
+  string network = 8;
+
+  // Capabilities to be added or removed. Capabilities are first removed then
+  // added.
+  message Capabilities {
+    repeated string add = 1;
+    repeated string remove = 2;
+  }
+
+  // Add or remove capabilities on this container. The list of valid
```
robshakir, line 365:
```suggestion
  // Add or remove capabilities to this container. The list of valid
```

maybe? Not sure.

robshakir (approved):

PR #207, opened 2024-08-26
Remove cert proto api

marcushines:


robshakir (commented):
I agree that it's right to remove this proto at this point, given the maturity of gNSI. 

One thing that some users might struggle with is that if they are using multiple gNOI services and are just pulling the latest release then they will not be able to absorb changes to other protos without removing their support for gNOI.Cert. Is that a problem?

PR #214, opened 2024-09-27
Update gNOI.bgp.proto with HARD_RESET and GRACEFUL_RESET

sachendras:
Current gNOI.BGP proto lacks guidance on how the ClearBGPNeighborRequest.Hard must be handled. As per RFC4486, code6 subcode 4,

```If a BGP speaker decides to administratively reset the peering with a neighbor, then the speaker SHOULD send a NOTIFICATION message with the Error Code Cease and the Error Subcode "Administrative Reset".``` 

which basically leaves implementations to make a choice if they should "reset TCP connection" and "Flush all routes OR not Flush all routes". Hence in rfc8538#section-3.2 for GR a new subcode 9 called `HARD RESET` was introduced to clear such ambiguities so the GR process can be handled more predictively.
Having said that, the gNOI.ClearBGPNeighborRequest API must also have 2 options for reseting a TCP connection to match the RFC. The `GRACEFUL` option introduced in this pull expects sending subcode 4 and the existing `HARD` option expects sending subcode 9 so the GR behavior can be accurately handled by expecting implementations to flush routes post reseting the TCP connection.

### DIFF
--- a/bgp/bgp.proto
+++ b/bgp/bgp.proto
@@ -39,12 +39,14 @@ message ClearBGPNeighborRequest {
   enum Mode {
     SOFT = 0;         // Send route-refresh and reapply policy.
     SOFTIN = 1;       // Re-apply inbound policy on stored Adj-RIB-In.
-    HARD = 2;         // Teardown, restart TCP connection, Send Cease 
+    HARD = 2;         // Teardown and restart TCP connection.
+    HARD_RESET = 3;         // Teardown, restart TCP connection, Send Cease 
     // notification (code 6) with subcode 4. However, if the "N" bit for 
     // Graceful-Restart" is set, then encapsulate the original Notification
     // message in the data portion of the Subcode 9 message as per 
     // rfc8538#section-3.1 and also Flush the routes.
-    GRACEFUL = 4;     // Teardown, restart TCP connection, Send Cease notification (code 6) with subcode 4.
+    GRACEFUL_RESET = 4;     // Teardown, restart TCP connection, Send Cease 
+    // notification (code 6) with subcode 4.
   }
   Mode mode = 3;       // Which mode to use for the clear operation.
 }
--- a/bgp/bgp.proto
+++ b/bgp/bgp.proto
@@ -40,7 +40,7 @@ message ClearBGPNeighborRequest {
     SOFT = 0;         // Send route-refresh and reapply policy.
     SOFTIN = 1;       // Re-apply inbound policy on stored Adj-RIB-In.
     HARD = 2;         // Teardown, restart TCP connection, Send Cease notification (code 6) with subcode 9 and flush all routes.
-    GRACEFUL = 4;     // Teardown, restart TCP connection, Send Cease notification (code 6) with subcode 4 and dont flush routes.
+    GRACEFUL = 4;     // Teardown, restart TCP connection, Send Cease notification (code 6) with subcode 4.
   }
   Mode mode = 3;       // Which mode to use for the clear operation.
 }
--- a/bgp/bgp.proto
+++ b/bgp/bgp.proto
@@ -40,13 +40,15 @@ message ClearBGPNeighborRequest {
     SOFT = 0;         // Send route-refresh and reapply policy.
     SOFTIN = 1;       // Re-apply inbound policy on stored Adj-RIB-In.
     HARD = 2;         // Teardown and restart TCP connection.
-    HARD_RESET = 3;         // Teardown, restart TCP connection, Send Cease 
-    // notification (code 6) with subcode 4. However, if the "N" bit for 
-    // Graceful-Restart" is set, then encapsulate the original Notification
-    // message in the data portion of the Subcode 9 message as per 
-    // rfc8538#section-3.1 and also Flush the routes.
-    GRACEFUL_RESET = 4;     // Teardown, restart TCP connection, Send Cease 
-    // notification (code 6) with subcode 4.
+    // If the "N" bit for Graceful-Restart" is set, then encapsulate the original 
+    // Notification message in the data portion of the Subcode 9 message per 
+    // rfc8538#section-3.1 and Flush routes associated with this neighbor.
+    // Otherwise, send BGP Cease notification (code 6) with subcode 4 then 
+    // teardown and restart the TCP connection.
+    HARD_RESET = 3;
+    // Send BGP Cease notification (code 6) with subcode 4 then 
+    // teardown and restart the TCP connection.
+    GRACEFUL_RESET = 4;     
   }
   Mode mode = 3;       // Which mode to use for the clear operation.
 }
--- a/bgp/bgp.proto
+++ b/bgp/bgp.proto
@@ -31,6 +31,11 @@ service BGP {
 
 }
 
+// ClearBGPNeighbor clears a BGP session. On success, this RPC returns an 
+// empty ClearBGPNeighborResponse. In case of a failure to meet a 
+// precondition (e.g., requesting a graceful reset on a session where 
+// Graceful Restart is not enabled), the server MUST return a gRPC error 
+// with the relevant status code.
 message ClearBGPNeighborRequest {
   string address = 1;  // IPv4 or IPv6 BGP neighbor address to clear.
   // Routing instance containing the neighbor. Defaults to the global routing
@@ -70,8 +75,8 @@ message ClearBGPNeighborRequest {
     // control plane restarts. The following actions are required for each scenario:
     
     // - If the BGP session does not support Graceful Restart (GR):
-    //  - The implementation must return an error, as a graceful reset cannot be
-    //   performed.
+    //  - The implementation MUST return a gRPC error with the status code
+    //    `FAILED_PRECONDITION`, as a graceful reset cannot be performed.
     
     // - If the BGP session supports GR but does not support RFC 8538:
     //   - The implementation must not send a BGP NOTIFICATION message.
--- a/bgp/bgp.proto
+++ b/bgp/bgp.proto
@@ -40,14 +40,54 @@ message ClearBGPNeighborRequest {
     SOFT = 0;         // Send route-refresh and reapply policy.
     SOFTIN = 1;       // Re-apply inbound policy on stored Adj-RIB-In.
     HARD = 2;         // Teardown and restart TCP connection.
-    // If the "N" bit for Graceful-Restart" is set, then encapsulate the original 
-    // Notification message in the data portion of the Subcode 9 message per 
-    // rfc8538#section-3.1 and Flush routes associated with this neighbor.
-    // Otherwise, send BGP Cease notification (code 6) with subcode 4 then 
-    // teardown and restart the TCP connection.
+    // When handling a request to perform a HARD_reset on a BGP session, the
+    // implementation's behavior is determined by the capabilities negotiated with
+    // the peer. The following actions are required for each scenario:
+    
+    // - If the BGP session does not support Graceful Restart (GR):
+    //     - The implementation must send a BGP Cease NOTIFICATION (code 6) to the
+    //       peer. The value of any subcode sent is considered irrelevant.
+    //     - It must reset the TCP session.
+    //     - It must flush all routes learned over this session.
+    
+    // - If the BGP session supports GR but does not support RFC 8538:
+    //     - The implementation must send a BGP Cease NOTIFICATION (code 6) to the
+    //       peer. The value of any subcode sent is considered irrelevant.
+    //     - It must reset the TCP session.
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: bgp/bgp.proto [hunk 1]
```diff
@@ -39,7 +39,12 @@ message ClearBGPNeighborRequest {
   enum Mode {
     SOFT = 0;         // Send route-refresh and reapply policy.
     SOFTIN = 1;       // Re-apply inbound policy on stored Adj-RIB-In.
-    HARD = 2;         // Teardown and restart TCP connection.
+    HARD = 2;         // Teardown, restart TCP connection, Send Cease
```
LimeHat, line 42: This is a (slightly) backward-incompatible change.
There could be implementations that send codes 4/9 unconditionally.

sachendras: Can you please elaborate this with examples? What does unconditionally mean? 
Error code 6 subcode 9 was introduced to handle GR cases when the "N" bit in the RESET flag of Capability 64 is set. So what other conditions would you send this subcode?

Following also from RFC8538 section#4:
`A BGP speaker SHOULD NOT send a Hard Reset to a peer from which it
   has not received the "N" bit.  We note, however, that if it did so,
   the effect would be as desired in any case because, according to
   [[RFC4271](https://datatracker.ietf.org/doc/html/rfc4271)] and [[RFC4724](https://datatracker.ietf.org/doc/html/rfc4724)], any NOTIFICATION message, whether recognized
   or not, results in a session reset.  Thus, the only negative effect
   to be expected from sending the Hard Reset to a peer that hasn't
   advertised compliance to this specification would be that the peer
   would be unable to properly log the associated information.`

LimeHat: Scenario 1) an implementation doesn't know about the RFC & code 9 and always uses code 4. (and receives the notification bit from its peer, which implies, according to the rules you specified, code 9)
Scenario 2) an implementation knows about the RFC and code 9 and still always uses code 4, because nobody said that it cannot do so for this specific RPC call
Scenario 3) an implementation knows about code 9 and always uses it, ignoring the quoted recommendation

There are a number of implementations in the wild, and any of these are possible. A safer way to introduce new (strict) requirements is to add a new option instead of redefining the existing one.

sachendras: Scenario 1) an implementation doesn't know about the RFC & code 9 and always uses code 4. (and receives the notification bit from its peer, which implies, according to the rules you specified, code 9)"

	Response: Right, which means the receiving speaker either isnt supporting RFC8538 or isnt configured for one. Expectations are that when the implementation does so, do follow the RFC requirement for "HARD RESET" when the gNOI.ClearBGPNeighborRequest.HARD is executed.

	Following from the RFC8538
	` A BGP speaker SHOULD NOT send a Hard Reset to a peer from which it has not received the "N" bit. We note, however, that if it did so, the effect would be as desired in any case because, according to [RFC4271] and [RFC4724], any NOTIFICATION message, whether recognized or not, results in a session reset.  Thus, the only negative effect to be expected from sending the Hard Reset to a peer that hasn't advertised compliance to this specification would be that the peer would be unable to properly log the associated information.`


Scenario 2) An implementation knows about the RFC and code 9 and still always uses code 4, because nobody said that it cannot do so for this specific RPC call.

	Response: Understood, which is why additional guidance now. For your example, following can happen
		a. The implementation doesnt support RFC8538 and hence hasnt set "N" bit in the initial Capability 64 negotiation. Therefore, the receipt of gNOI.ClearBGPNeighborRequest.HARD command would result in the implementation sending error code 6 subcode 4. This MUST result in TCP connection reset and route FLUSH. In this scenario, gNOI.ClearBGPNeighborRequest.GRACEFUL will also result in sending error code 6 subcode 4. Since bit "N" of the RESET flag is "0", the recieving speaker would handle the subcode4 receipt accordingly resulting in route flush.
		b. If the "N" bit was set originally then for "gNOI.ClearBGPNeighborRequest.HARD", the expectations are that the original Code6 Subcode 4 is encapsulated inside the errorcode 6 subcode9. If the implementation doesnt do so and continues to send error code 6 subcode 4, then this wont be a HARD_RESET signal for the receiving Speaker of the NOTIFICATION message. In this case the TCP session can go down but the routes wouldnt be flushed. This is also the RFC 8538 expectation for Subcode 4.

		Adding a 3rd option in the proto keeping "HARD" intact will be confusing. The issue is, the currently defined "HARD" wasnt comprehensive enough to cover for GR scenarios. This pull is intending to close that gap. 

Scenario 3) an implementation knows about code 9 and always uses it, ignoring the quoted recommendation
	Response: What does that mean? Subcode 9 was introduced in the RFC8538 for GR. If the implementation is using that all the time then they arent compliant with the original BGP CEASE message (rfc4486).

LimeHat: >Adding a 3rd option in the proto keeping "HARD" intact will be confusing. The issue is, the currently defined "HARD" wasnt comprehensive enough to cover for GR scenarios. This pull is intending to close that gap.

The previous option can be deprecated.

sachendras: Just for my knowledge, how is that approach of deprecating "HARD" in the favor of a new attribute to handle GR and non-GR situation for "HARD" reset better than updating the existing "HARD" attribute itself?

LimeHat: It doesn't change the end goal, the issue is just a matter of backward-compatibility.

As you correctly noted, in the current proto the behavior is not fully defined, so there could be differences in existing implementations (we can debate what is correct and what is not correct according to the rfc and so on, but that's not the main point).

A new option (for example let's call it `HARD_RFC8538`) provides an API guarantee to a client: if the client calls it, he gets the exact, strictly defined behavior in all cases (subcode 9 if supported by the receiver, subcode 4 otherwise). 

Updating the existing option doesn't provide such guarantees: the client will get this behavior if NOS was updated according to the latest changes in proto, but if it wasn't, the behavior is not really deterministic and maybe the NOS will continue sending subcode 4 while the gnoi client wanted to send code 9.

sachendras: Added the following instead so hopefully that addresses the concern here

HARD_RESET = 3;         // Teardown, restart TCP connection, Send Cease 
    // notification (code 6) with subcode 4. However, if the "N" bit for 
    // Graceful-Restart" is set, then encapsulate the original Notification
    // message in the data portion of the Subcode 9 message as per 
    // rfc8538#section-3.1 and also Flush the routes.

GRACEFUL_RESET = 4;     // Teardown, restart TCP connection, Send Cease 
    // notification (code 6) with subcode 4.

bgp/bgp.proto
```diff
@@ -40,6 +40,13 @@ message ClearBGPNeighborRequest {
     SOFT = 0;         // Send route-refresh and reapply policy.
     SOFTIN = 1;       // Re-apply inbound policy on stored Adj-RIB-In.
     HARD = 2;         // Teardown and restart TCP connection.
+    HARD_RESET = 3;         // Teardown, restart TCP connection, Send Cease 
+    // notification (code 6) with subcode 4. However, if the "N" bit for 
+    // Graceful-Restart" is set, then encapsulate the original Notification
+    // message in the data portion of the Subcode 9 message as per 
+    // rfc8538#section-3.1 and also Flush the routes.
+    GRACEFUL_RESET = 4;     // Teardown, restart TCP connection, Send Cease 
+    // notification (code 6) with subcode 4.
```
dplore, line 49: Formatting and ordering change for the description

```suggestion
    // If the "N" bit for Graceful-Restart" is set, then encapsulate the original 
    // Notification message in the data portion of the Subcode 9 message per 
    // rfc8538#section-3.1 and Flush routes associated with this neighbor.
    // Otherwise, send BGP Cease notification (code 6) with subcode 4 then 
    // teardown and restart the TCP connection.
    HARD_RESET = 3;
    // Send BGP Cease notification (code 6) with subcode 4 then 
    // teardown and restart the TCP connection.
    GRACEFUL_RESET = 4;     
```

dplore: It seems we want two things:

1. Hard reset the connection and override graceful restart, if it is enabled
2. Hard reset the TCP connection, but maintain graceful restart, if it is enabled.

The current field (`HARD = 2`) is ambiguous in the presence of graceful restart.  I suggest we deprecate it.  How does that sound?

How about we name the two new fields like this:

`HARD_RESET_GRACEFUL` = 3    // Reset and restart the TCP connection, preserving routes if  graceful restart is enabled
`HARD_RESET_GRACEFUL_OVERRIDE` = 3   // Reset and restart the TCP connection, overriding graceful restart per RFC8538 Section 3.1


This avoids restating or interpreting the relevant RFC (which I agree is a little difficult to read, so I see why it's tempting to clarify it here)

bgp/bgp.proto
```diff
@@ -40,6 +40,15 @@ message ClearBGPNeighborRequest {
...
     SOFTIN = 1;       // Re-apply inbound policy on stored Adj-RIB-In.
     HARD = 2;         // Teardown and restart TCP connection.
+    // If the "N" bit for Graceful-Restart" is set, then encapsulate the original 
+    // Notification message in the data portion of the Subcode 9 message per 
+    // rfc8538#section-3.1 and Flush routes associated with this neighbor.
+    // Otherwise, send BGP Cease notification (code 6) with subcode 4 then 
+    // teardown and restart the TCP connection.
+    HARD_RESET = 3;
+    // Send BGP Cease notification (code 6) with subcode 4 then 
+    // teardown and restart the TCP connection.
```
rszarecki, line 50:
```suggestion
     // If session do not support GR, then implementation shall return error
     // If session nedotiated capabilities support GR 
     // but do not support RFC 8538, then implementation shall:
     //   - Do NOT send NOTIFICATION
     //  - Reset TCP session
     //  - Retain all routes lerned over this session for period of time
     //    derived form restart-time, LLGR and/or ExRR depending
     //    on configuration and negotiated capabilities.
     // If session do support GR and RFC 8538, 
     //  then implementation shall:
     //   - Send NOTIFICATION to peer BGP Cease notification (code 6)
     //   with subcode "Other Configuration Change" (value 6)
     //  - Reset TCP session
     //  - Retain all routes lerned over this session for period of time
     //    derived form restart-time, LLGR and/or ExRR depending
     //    on configuration and negotiated capabilities.
```

bgp/bgp.proto
rszarecki, line 1: I think
```
message ClearBGPNeighborResponse {
}
```
Need to be extended to provide Error Codes:

"SUCCESS"
"UNSUPPORTED_MODE"
"OTHER_ERROR"

bgp/bgp.proto
```diff
@@ -40,6 +40,15 @@ message ClearBGPNeighborRequest {
     SOFT = 0;         // Send route-refresh and reapply policy.
     SOFTIN = 1;       // Re-apply inbound policy on stored Adj-RIB-In.
     HARD = 2;         // Teardown and restart TCP connection.
+    // If the "N" bit for Graceful-Restart" is set, then encapsulate the original 
+    // Notification message in the data portion of the Subcode 9 message per 
+    // rfc8538#section-3.1 and Flush routes associated with this neighbor.
+    // Otherwise, send BGP Cease notification (code 6) with subcode 4 then 
+    // teardown and restart the TCP connection.
```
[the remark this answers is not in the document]
sachendras: Content included, thanks!

bgp/bgp.proto
sachendras: Content included, thanks!

bgp/bgp.proto
sachendras: I added comments to include a gRPC error message in response when the precondition isn't meat.

dplore:
Lint check is failing:
https://github.com/openconfig/gnoi/actions/runs/15863949225/job/44727075118?pr=214

bgp/bgp.proto
```diff
@@ -40,6 +45,55 @@ message ClearBGPNeighborRequest {
     SOFT = 0;         // Send route-refresh and reapply policy.
     SOFTIN = 1;       // Re-apply inbound policy on stored Adj-RIB-In.
     HARD = 2;         // Teardown and restart TCP connection.
```
dplore, line 47: Deprecate the HARD field and add description such as:

```suggestion
    // The HARD Mode is intended to teardown and restart the BGP session
    // TCP connection. This field does not define what the behavior is regarding
    // BGP graceful restart. The HARD_RESET and GRACEFUL_RESET fields 
    // should be used instead of HARD.
    HARD = 2;      [deprecated=true];
```

dplore (changes requested):

ElodinLaarz:
Talked about this at the July 1st meet -- Seems reasonable.

dplore:
FYI:  Last call deadline is July 8th, 2025, then this will be merged.

PR #225, opened 2024-11-07
Adding metadata to container listing

alshabib:
This commit adds the following:

* ListResponse must now return any metadata attached to a container.
* ListResponse must now return the container hash.
* RemoveContainer should remove the specified container


containerz/containerz.proto
```diff
@@ -54,12 +54,10 @@ service Containerz {
...
-  // RemoveContainer deletes images that match the spec defined in the
-  // request. If the image is associated to a running container then an error
-  // is returned. If the specified container image does not exist, this
-  // operation is a no-op.
-  //
-  // Deprecated - use RemoveImage instead.
+  // RemoveContainer deletes the container that matches the instance
+  // name. If the specified container does not exist, this
+  // operation is a no-op. The container should not be removed if it is
+  // running unless force is specified in the request.
```
robshakir, line 60: We should probably update the PR description to say that the container should not be removed if it's running -- I came to look here since the description says something like "`RemoveContainer` should not remove a container"!

containerz/containerz.proto
```diff
@@ -225,11 +223,10 @@ message RemoveImageResponse {
 }
 
 message RemoveContainerRequest {
-  // The container image name to be removed.
-  string name = 1;
+  // The container name to be removed.
+  string instance_name = 1;
```
robshakir, line 227: We should not just change the field name her -- any textproto instances will no longer unmarshal correctly. Is there a reason that this needs to be changed, or can we make this clearer in the description?

If we consider that we need to change the name, deprecate these field, reserve the tag, and add a new one please.

containerz/containerz.proto
```diff
@@ -314,6 +311,12 @@ message ListContainerResponse {
   }
 
   Status status = 4;
+
+  // Metadata associated to this container.
+  map<string, string> labels = 5;
+
+  // The computed hash as returned by the container runtime.
```
robshakir, line 318: The computed hash of the image?

containerz/containerz.proto
```diff
@@ -314,6 +311,12 @@ message ListContainerResponse {
   }
 
   Status status = 4;
+
+  // Metadata associated to this container.
+  map<string, string> labels = 5;
```
robshakir, line 316: Do we need to define something about what is expected to be returned here? Is this any metadata? Can I pack arbitrary things into it as the server?

(I'm always a bit skeptical of these 'unstructured' ways to add data to an API, since there's many different things one _can_ put into them, and it becomes unclear what one really expects to get back.)

containerz/containerz.proto
alshabib: Oh oops - I dropped the `not`

containerz/containerz.proto
alshabib: Ah good point about textprotos, I will revert this change as it is not entirely necessary.

containerz/containerz.proto
alshabib: yes - updated.

containerz/containerz.proto
alshabib: This is about associating the running container with some labels in the form of arbitrary key-values. So for example, commit the binary was built at or version of the application. We could be prescriptive about these keys but that could lead to adding a bunch of fields here.

robshakir: This makes sense -- can we (as discussed) just note that this should not change the behaviour of the RPCs so that we ensure that we don't end up with this being used to avoid adding new fields to the proto.

robshakir (approved):

robshakir:
Looks like `staticcheck` is unhappy with generated `.pb.go` -- not sure if this can be fixed by just running `protoc-gen-go`.

robshakir:
Looks like kicking the bazel cache solved that issue -- just the staticcheck one to look at,

alshabib:
> Looks like kicking the bazel cache solved that issue -- just the staticcheck one to look at,

Not sure why the static check is failing here - I regenerated the pb.go files and I still get the error. Does this break for you too or is it just me?

.github/workflows/go.yml
```diff
@@ -10,3 +10,5 @@ on:
 jobs:
   go:
     uses: openconfig/common-ci/.github/workflows/basic_go.yml@v0.2.0
+    with:
+      skip-staticcheck: true
```
robshakir, line 14: Let's not do this.

robshakir (changes requested):
Blocking merge whilst we fix staticcheck in a more robust way.

robshakir:
Can you pull `main` into this branch? It doesn't seem to give me the option to do so.

alshabib:
Updated - thank you!

robshakir (approved):

Issue #149, opened 2023-11-07
Clarify PLQ capabilities fields


LimeHat:
Currently `PacketGeneratorCapabilities` message does not explicitly define `max_bps` and `max_pps` fields usage. Are they supposed to indicate capabilities of a single ASIC? What if there are multiple ASICs with different capabilities in the system?

Let's say we have a modular system with 
* 2x linecards (LC1 and LC2) with different ASICs, 2x ASICs in each LC
* ASICs in LC1 are capable of generating 150 Gbps of traffic each
* ASICs in LC2 are capable of generating 750 Gbps of traffic each

Should the system return 750 Gbps in `max_bps` field of `PacketGeneratorCapabilities` message?

*same considerations would be applicable to mtu fields as well

PR #228, opened 2024-12-06
Add support for PBLQ per-interface capabilities

earies:
Physical/channelized interfaces could support a variety of Generator or
Reflector capabilities.  An optional field is added to each capability
to report a list of supported interfaces.  Should this field not be set,
it is assumed that all interfaces are supported for the capability
reported.


packet_link_qualification/packet_link_qualification.proto
```diff
@@ -326,6 +331,9 @@ message PacketInjectorCapabilities {
   // Loopback modes that are supported by the device when using the packet
   // injector mode.
   repeated PacketInjectorLoopbackMode loopback_modes = 8;
+  // List of interfaces that support this capability.  If this field is not
+  // populated then it is assumed that all interfaces are supported.
+  repeated string supported_interfaces = 9;
```
anshkhanna-ggl, line 336: Not sure if this makes sense but an interface can have multiple capabilities right?

earies: The proto structs currently assumes system wide

Each capability is expanded in this patch set to accomodate a list of interfaces per capability so yes, the same interface could be reported under each capability if an interface supports multiple capabilities.

Another option is to create an entirely new capabilities API (or modify existing in non-backwards compatible) that does not make the system wide assumption from the start and allows for further granularity

LimeHat:
While I agree that a system-wide capabilities response is not always sufficient (I raised some of the questoins/issues in #149); I'm not a huge fan of per-interface approach. Typically in HW capabilities are defined per ASIC; I'd suggest considering a per-ASIC capabilities response format instead.

earies:
> While I agree that a system-wide capabilities response is not always sufficient (I raised some of the questoins/issues in #149); I'm not a huge fan of per-interface approach. Typically in HW capabilities are defined per ASIC; I'd suggest considering a per-ASIC capabilities response format instead.

Since this API deals with "links" (e.g. interfaces), the dimension of an ASIC seems that it would unfold into needing yet another method to resolve what the ASIC -> interface relationship is (either within this API or a dependency on another API)

We also have a use-cases that aren't distinguishable by an ASIC alone but rather configuration attributes of an interface that cascades towards separate capablities "per interface"

LimeHat:
>We also have a use-cases that aren't distinguishable by an ASIC alone but rather configuration attributes of an interface that cascades towards separate capablities "per interface"

Can you please elaborate on this part?

>Since this API deals with "links" (e.g. interfaces), the dimension of an ASIC seems that it would unfold into needing yet another method to resolve what the ASIC -> interface relationship is (either within this API or a dependency on another API)

That is a fair point. gNMI/OC can be used to establish that relationship if necessary via the components tree.
Today we already have indirect dependencies on another API - e.g. PLQ clients have to know what interfaces are available in the system, interface speeds, etc. Unless the objective of your PR is to eliminate all of that, I don't see much of a problem with extending this approach to asic-to-interface mappings.

earies:
> > We also have a use-cases that aren't distinguishable by an ASIC alone but rather configuration attributes of an interface that cascades towards separate capablities "per interface"
> 
> Can you please elaborate on this part?

One example could be the modes set on an interface (e.g. channelization) - if we get into implementation detail, there could be a variety of reasons that capabilities can then change for the full intent of these APIs.  We don't have this level of granularity in the API definition today to signal any differences as there is a system wide assumption.

> > Since this API deals with "links" (e.g. interfaces), the dimension of an ASIC seems that it would unfold into needing yet another method to resolve what the ASIC -> interface relationship is (either within this API or a dependency on another API)
> 
> That is a fair point. gNMI/OC can be used to establish that relationship if necessary via the components tree. Today we already have indirect dependencies on another API - e.g. PLQ clients have to know what interfaces are available in the system, interface speeds, etc. Unless the objective of your PR is to eliminate all of that, I don't see much of a problem with extending this approach to asic-to-interface mappings.

Personally, I think this is also something that is not well understood or prescribed from an overarching OC umbrella architectural standpoint.  It is often where intent or data is duplicated amongst APIs or relationships that are constructed that assume OC YANG schemas for what could (possibly should?) be schema agnostic APIs.  It isn't until a workflow is really dissected before realizing that there may need to be a dependence on another interface to access pre-requisite data.  In this case, I had also thought we _could_ convey such capabilities in a YANG state subtree somewhere as well so it begs the overall question, should you be able to fulfill PBLQ related intent solely w/ the APIs defined within this scope?  To your point, we already have a dependence on an alternate API here for prior context (likely gNMI subscribe/get but also not limited to depending)

The objective of this PR is primarily to provide a mechanism to signal potentially different capabilities per "link" and a client handling this prior to further RPC execution.  How that is accomplished is best suited for the most optimal workflow - I don't feel strongly whether it should exist in this API or elsewhere tbh...

earies:
cc: @dplore @robshakir @anshkhanna-ggl for additional thoughts on some of the discussion here

LimeHat:
>One example could be the modes set on an interface (e.g. channelization) - if we get into implementation detail, there could be a variety of reasons that capabilities can then change for the full intent of these APIs.

Can you provide specific examples (what capabilities are expected to have per-intf granularity)? I'm not sure if I'm following. 

Thanks!

PR #234, opened 2024-12-15
Adding limits to container start operations

alshabib:


containerz/containerz.proto
```diff
@@ -417,6 +417,30 @@ message StartContainerRequest {
   RunAs run_as = 11;
 
   map<string, string> labels = 12;
+  
+  // Limits defines restrictions or limits upon this container.
+  message Limits {
+    // max_cpu specifies the maximum amount of cpu this container can use.
+    // For example a value of 0.5 means this container can use a maximum of
+    // half a cpu. This setting does not imply that the container is pinned
```
robshakir, line 425: Is this half a CPU or half a core? (I assume that this means 50% of a core that is exposed to the kernel -- which I think is referred to as a CPU in some commands, e.g., `top(1)`.) k8s uses the definition:

> 1 Hyperthread on a bare-metal Intel processor with Hyperthreading [ref](https://kubernetes.io/docs/tasks/configure-pod-container/assign-cpu-resource/)

I presume that this is going to some underlying cgroups API eventually, so the definition could come from there.

containerz/containerz.proto
```diff
@@ -417,6 +417,30 @@ message StartContainerRequest {
...
+    // to a cpu but rather only the maximum amount of cpu it can use.
+    double max_cpu = 1;
+
+    // soft_mem defines a soft memory limit. If the system is not memory
+    // constrained this limit has no effect. If it is memory constrained
+    // this container may be oom killed if it exceed this limit. This value
+    // is effectively a soft memory limit. This limit should be smaller than
+    //  max_mem otherwise it will have no effect.
+    int64 soft_mem =2;
```
robshakir, line 434: What are the units here? Generally, I think best practice is to have the units in the name -- e.g., `soft_mem_bytes`.

containerz/containerz.proto
```diff
@@ -417,6 +417,30 @@ message StartContainerRequest {
...
+
+    // hard_mem is a hard limit. If the container exceed this number the 
+    // container runtime can take aggressive action such as terminate (oom kill)
+    // the container.
+    int64 hard_mem = 3;
```
robshakir, line 439: Same question above re: units.

robshakir (changes requested):

containerz/containerz.proto
alshabib: Hmm googd point - I think we should define this as a CPU and the specific implementation for the container runtime can translate accordingly.

containerz/containerz.proto
alshabib: ah yes - thanks!

containerz/containerz.proto
alshabib: Thanks!

containerz/containerz.proto
robshakir: I think we can just say "an OS-visible CPU (which may map to a hyperthread on an Intel processor with hyperthreading, or a CPU core".

robshakir (approved):

PR #237, opened 2024-12-18
Bazel7 Workspace Support

bstoll:
This change updates the build process to use newer tooling with Bazel WORKSPACEs.

* In WORKSPACE, prefer to load dependencies closest to the project first.
* Gazelle builds go deps (`bazel run //:gazelle_update_repos`)
* Gazelle can generate BUILD files (`bazel run //:gazelle`) for protos & Go.
* Sync up go.mod and WORKSPACE versions
* Start using import_prefixes more consistently. This should make GRPC reflection work a bit better.

This is notably missing bzlmod support. We need to import GNSI, GNMI, and bootz into this repo but none of these use bzlmod yet. I believe the best way is to start at the bottom and migrate them one by one (starting with GNMI) - alternatively, we could patch them up but this is a bit clunky to do.

The GRPC compiler has an arg difference that results in different generated output: GRPC interfaces generally include mustEmbedUnimplementedXYZ, while the @io_bazel_rules_go//proto:go_grpc_v2 does not enforce this. Since we already generate with the mustEmbed, it would seem ideal to keep doing this.


bstoll:
Closing this in favor of #248.

PR #238, opened 2024-12-19
adding support for plugins

alshabib:


containerz/containerz.proto
```diff
@@ -118,6 +119,23 @@ service Containerz {
 
   // ListVolume returns all volumes that match the spec defined in the filter.
   rpc ListVolume(ListVolumeRequest) returns (stream ListVolumeResponse) {}
+
+  // StartPlugin starts the specified plugin identified by the name field
+  // of the request. It should be started using the instance_name of the
+  // request. If the plugin file named `name.tar` does not exist, this
```
robshakir, line 125: I presume the plugin needs to be a packaged OCI image right? Does this necessarily mean it must be `name.tar`?

containerz/containerz.proto
```diff
@@ -118,6 +119,23 @@ service Containerz {
...
+  // should return an error.
+  rpc StartPlugin(StartPluginRequest) returns (StartPluginResponse) {}
```
robshakir, line 127: Given we have separate plugin-specific commands for start/stop/list -- did you consider having a different one for `Deploy` to make the API consistent? (Not sure of the pros/cons.)

Is there a link to how volume plugins look for podman? Best I could find is https://github.com/containers/podman/discussions/23762.

containerz/containerz.proto
alshabib: Yes - fair point. Basically we need to make sure that the file uploaded during deploy is referenceable when PluginStart is called. So there is no need to mandate an extension.

containerz/containerz.proto
alshabib: AFAIK Docker plugins can be used in podman so podman plugins are the same as docker plugins. 

I thought about a separate Deploy API but it seems like a lot of duplication since Deploy already uploads a file as we need.

containerz/containerz.proto
```diff
@@ -39,9 +39,10 @@ service Containerz {
   // Deploy sets a container image on the target. The container is sent as
   // a sequential stream of messages containing up to 4MB of data (max default
   // gRPC message size). Upon reception of a valid container, the target must
-  // load it into its registry. Whether the registry is local or remote is
-  // target and deployment specific. A valid container is one that has passed
-  // its checksum.
+  // load it into its registry, unless the file is a plugin. If the file is a 
+  // plugin, the target must save the file until PluginStart is called. Whether
```
robshakir, line 43: `s/PluginStart/StartPlugin/g`.

containerz/containerz.proto
robshakir: Sounds reasonable to me -- I was forgetting that `Deploy` didn't actually start the container, so this seems reasonable to me.

containerz/containerz.proto
```diff
@@ -144,6 +162,9 @@ message ImageTransfer {
   // Optional. Instructs the target to fetch the image from a remote location.
   // The above name and tag must be applied to the container once downloaded.
   common.RemoteDownload remote_download = 4;
+
+  // Indicates that the file to be transferred is a plugin.
+  bool is_plugin = 5;
```
robshakir, line 167: Should this be within a `oneof` or alternatively an enum? It seems like we might have some other types that are mutually exclusive that might be transferred via `Deploy`?

alshabib: That's not clear to me - I don't think there are other common types images. Technically, this only tells the server not to attempt to load the received image so perhaps it should be `no_load` instead of `is_plugin`

PR #244, opened 2025-01-06
Add ACL Policy for Debug service. This will enable the support of

marcushines:
per user per command ACL for the service.

this will allow the service to act as a grpc based replacement for ssh access that was previously acl'ed via username to tacacs


debug/debug.proto
```diff
@@ -75,3 +75,20 @@ message DebugStatus {
   // A list of messages that carry the error details.
   repeated google.protobuf.Any details = 3;
 }
+
+// Policy defines a set of commands a list of users are allowed to execute
```
robshakir, line 79: How is this provided to the device?

It seems a little odd that we don't have this in gNSI -- is there some opportunity to express this there akin to pathz?

marcushines: I can move it into gnsi as far as providing the gnsi.Debug.Rotate it just kinda feels a bit odd if that is really going to be a pattern we follow for all services

marcushines: are you also suggesting trying to make pathz support other services? I am not sure it is worth trying to fit that - "role based access control" generally is tailored to the service implementation conversely the definition of Authz is generic only to grpc itself

PR #254, opened 2025-01-23
Adding options for custom volume plugins

alshabib:


containerz/containerz.proto
```diff
@@ -629,14 +634,17 @@ message CreateVolumeRequest {
   // and associated options.
   Driver driver = 2;
 
-  // Options for the driver. The actual option keys and values are driver
+  reserved 3;
+
+  // Labels to apply to the volume. Labels are metadata for the the volume.
+  map<string, string> labels = 10;
+
+  // Options for the driver. The actu:al option keys and values are driver
```
robshakir, line 642:
```suggestion
  // Options for the driver. The actual option keys and values are driver
```

containerz/containerz.proto
```diff
@@ -629,14 +634,17 @@ message CreateVolumeRequest {
...
   // specific.
   oneof options {
-    LocalDriverOptions local_mount_options = 3;
+    LocalDriverOptions local_mount_options = 20;
+    CustomOptions custom_options = 21;
   }
-
-  // Labels to apply to the volume. Labels are metadata for the the volume.
-  map<string, string> labels = 4;
```
robshakir, line 639: Any reason that this changed field ID but looks to have stayed in the same place in the proto?

alshabib: Well I wanted to create some room in the proto for future fields and options so I figured this is a better time to pay this price rather that later when this is more in use.

robshakir: Not sure I understand -- we're making a backwards incompatible change, but I don't understand why. Maybe missing something.

If we have `labels` continue to use tag 3, it's not like we can't add other fields later on elsewhere -- the same is true for the tag change for `local_mount_options`. The tags don't have to be sequential in the proto.

alshabib: Unfortunately I had sandwiched options between driver (2) and labels (4). So I can either do what I did in this PR or other options would have to start at 5 (or more). Happy to do either I just opted for the one which would be less error prone going forward.

robshakir: IMHO, we should bias towards backwards compatibility as soon as these things are committed and consumed. Any proto that evolves ends up having some of these out of order  field tag issues.

alshabib: True - I restablished the numbering.

containerz/containerz.proto
```diff
@@ -619,6 +619,11 @@ message LocalDriverOptions {
   string mountpoint = 3;
 }
 
+message CustomOptions {
```
robshakir, line 622: Should we describe here what the "custom" driver is? Is there a documentation reference point for this?

(I'm also curious whether this 'custom' driver is something that we're expecting to be supported across different container runtimes.)

alshabib: There is no one custom driver - this set of options applies to any third party volume plugin. The container manager simply passes these options to the plugin/driver.

robshakir: LGTM, thanks for adding the comment.

robshakir (approved):

PR #268, opened 2025-03-21
clarifying how errors should be returned during remove container oper…

alshabib:
clarifying how errors should be returned during remove container operations


containerz/containerz.proto
```diff
@@ -52,29 +52,29 @@ service Containerz {
...
   // Start starts a container. If the image does not exist on the target,
   // Start returns an error. A started container is identified by an instance
   // name, which  can optionally be supplied by the caller otherwise the target
-  // should provide one. When the instance name already exists, the target 
-  // should return an error if the container is running or if it is stopped but 
-  // the request has more fields than just the instance name. If the container 
-  // is stopped and the request contains only the instance name then Start 
-  // should start the container
+  // should provide one. If the instance name already exists, the target should
+  // return an error.
```
robshakir, line 77: Is there a specification here as to what error code should be returned?

containerz/containerz.proto
```diff
@@ -359,31 +342,6 @@ message Volume {
   bool read_only = 3;
 }
 
-// A device to be added to the container.
```
robshakir, line 362: Is your client up-to-date? It looks like this was what was added in #264 being removed.

containerz/containerz.proto
```diff
@@ -655,14 +604,6 @@ message LocalDriverOptions {
   string mountpoint = 3;
 }
 
-// CustomOptions are an opaque set of volume options that are passed to the
```
robshakir, line 658: Is this removal intentional?

robshakir (changes requested):

containerz/containerz.proto
alshabib: Thanks I added one

containerz/containerz.proto
alshabib: oops good catch

containerz/containerz.proto
alshabib: Not it is not - thanks for flagging

containerz/containerz.proto
```diff
@@ -53,28 +61,33 @@ service Containerz {
...
   // Start starts a container. If the image does not exist on the target,
   // Start returns an error. A started container is identified by an instance
   // name, which  can optionally be supplied by the caller otherwise the target
-  // should provide one. When the instance name already exists, the target 
-  // should return an error if the container is running or if it is stopped but 
-  // the request has more fields than just the instance name. If the container 
-  // is stopped and the request contains only the instance name then Start 
-  // should start the container
+  // should provide one. When the instance name already exists, the target
+  // should return an error if the container is running or if it is stopped but
```
robshakir, line 86: I think you lost a change to specify the error type here.

alshabib: Ah yes thank you.

robshakir (approved):

robshakir:
LGTM, can you bump the version before submitting?

PR #294, opened 2025-08-07
factory_reset: add mechanism to delete images

brianneville:
Add a way to delete images on the system before proceeding with the factory reset.


brianneville:
superseded by https://github.com/openconfig/gnoi/pull/301

PR #309, opened 2025-09-19
Introduce a BootzConfiguration field to FactoryReset.Start().

gmacf:
## Overview

This PR introduces a new field to the gNOI FactoryReset.Start() gRPC that allows the caller to initiate Bootz without the need for an external DHCP server. If this field is populated, the device will factory reset as normal, but skip the process of acquring a DHCP lease.

### Separation of concerns

When a Bootz integration test fails, it's difficult to know if the DHCP part is failing, the Bootz part, or both. By separating these, we can individually test each component. 

### Testbed infrastructure simplicity

A Bootz test would no longer require a dedicated DHCP server in the lab, easing requirements on the testbed infrastructure.  For large lab setups, this often requires changes to routing configurations and the addition of DHCP relay agents, all of which add to its complexity. 

### Virtualized environment challenges

This approach bypasses some additional challenges with running integration tests in virtualized environments such as Kubernetes where users don't have as much control over L2 networking.


dplore:
We'll close #204 as stale.   

To further separate concerns for the device implementation side of this effort, maybe a new service such as  gnoi.BootzTest.<rpc for each testing verb> is suitable.    Are there other bootz testing actions you would want to take aside from a "fake DHCP response"?     For the use case you've outlined, perhaps your test client would send a  `gnoi.BootzTest.StartBootstrap(BootstrapResponseData)` to the device?

PR #317, opened 2025-11-24
updating containerz error handling

alshabib:


containerz/containerz.proto
```diff
@@ -60,17 +60,25 @@ service Containerz {
...
   // RemoveContainer deletes the container that matches the instance
   // name. If the specified container does not exist, this
   // operation is a no-op. The container should not be removed if it is
   // running unless force is specified in the request.
-  // The request must return an error with status code NotFound if the container
-  // is not found. If the container is running and the request did not set the
-  // force field to true then the request must return an error with status code
-  // FailedPrecondition.
+  //
+  // This RPC ill return the following error codes in case of error:
```
robshakir:
```suggestion
  // This RPC returns the following error codes in case of error:
```

robshakir (commented):
Minor comments.

I assume that this should also change the version of containerz since there'll be different implementation behaviours (potentially?).

robshakir:
:shipit:

PR #320, opened 2025-12-05
Specify error codes for gnoi.File

dplore:
It's observed that implementations vary in which error codes are used when files or directories do not exist.  This PR clarifies that the gRPC error code `NOT_FOUND` must be used for these conditions.  It also provides a suggestion for mapping file errors to gRPC errors for other common error conditions as a reminder to implementors.


dplore:
@ElodinLaarz  please review!

PR #327, opened 2026-02-12
gnoi/os: add functionality to download from remote

brianneville:
add a field enabling the client to instruct the
target to perform a remote download of the
OS package, rather than requiring the client
to stream the content to the server.


os/os.proto
```diff
@@ -161,6 +175,9 @@ message InstallRequest {
     TransferRequest transfer_request = 1;
     bytes transfer_content = 2;
     TransferEnd transfer_end = 3;
+
+    // Details for the device to download the image from a remote location
+    common.RemoteDownload remote_download = 4;
```
marcushines: i would have worries about the integrity of this OS download
seems like there would need to be a better way to guarantee this is a valid image before progressing

brianneville: The server will only be able to validate its integrity after it has completed the download, similar to how this validation works for images sent over in `transfer_content` by the client. 
If the validation succeeds, the server can return a `Validated` response, or if it fails it can return an `InstallError`.

brianneville:
bump - @marcushines  any further comments on this PR?

marcushines (approved):

PR #355, opened 2026-08-30
healthz: Add initiate_only field to CheckRequest

ejbrever:
### Description
Adds the `initiate_only` boolean field to `CheckRequest` in `healthz.proto`, bumps the `gnoi_version` to 1.4.0, and regenerates `healthz.pb.go`.

### Motivation
Allows clients to initiate an asynchronous health check on a specific component path without blocking the RPC until the check or diagnostic artifact generation has finished. Clients can subsequently query results via `Get` or `List`.


ejbrever:
@dplore @robshakir could you please review?

dplore:
This looks good to me in concept.  Do we need something to indicate a check is in progress?  (add IN_PROGRESS to the Status field?)

dplore:
Added to Sep 22, 2026 OC Operator review

ejbrever:
> This looks good to me in concept. Do we need something to indicate a check is in progress? (add IN_PROGRESS to the Status field?)

That's a good idea. I think the Status enum as-is describes the component health, so I thought it might be preferable to add a "collection_state" instead that is explicit for this. I made this change, but please let me know your thoughts.

